PR #16, opened 2026-08-29
Come back to the row a folder was left on

### LINKED ISSUE #6: Come back to where you were in a folder you have already been in
### What are you trying to do

Leaving a folder loses your place in it. Go into Amiga, get down to Super
Frog, go back out to look at something else, come back: the list is at the
top again and the scroll down happens a second time.

It is a small thing once. On a card where a folder holds thousands of games
it is the difference between browsing and starting over, and it happens
every time you look at two systems in a row.

Degauss already comes back to the right game after playing one, so it
already reads as something that remembers where you were. It just does not
do it when you leave a folder without launching anything, which is most of
the time.

### What you have in mind

Each list keeps the place it was left at, and restores it when you come back.

Most of this exists. `state.rs` already has a `SavedPlace`, a `trail` of
them with a `selected` for each, and the loading and saving. It is written
when a game is launched, so the trail comes back with you after playing, and
consumed once. What is missing is keeping the same thing live while
browsing, rather than only across a launch.

Three things worth deciding rather than discovering:

**How long it lasts.** Within one run of Degauss is the smaller change and
covers the case above. Across restarts is nicer and means `state.toml`
holding a table keyed by folder instead of one trail.

**How many are kept.** One entry per folder visited, on a card with a
hundred and more systems and thousands of folders, grows without a bound.
A cap, oldest dropped, is probably enough, and only matters if it survives
restarts.

**What happens when the folder changed.** Leaving on Super Frog, adding
games and rebuilding means the row that was at that index is a different
game. Remembering a number would land somewhere arbitrary, which is worse
than going back to the top, because it looks like the list moved on its own.
It has to remember what was selected by name and find it again, and fall
back to the top when it is gone. `SavedPlace` already works this way for the
resume path, so the pattern to follow is the one already there.

Done when:

- [ ] leaving a folder keeps its place, and returning restores it
- [ ] a folder whose contents changed restores by name, or goes to the top
- [ ] whether it survives a restart, decided and documented
- [ ] a bound on how many are kept, if it does
- [ ] the existing resume after a launch keeps working

### What you tried instead

The letter grid, to get back down to roughly where you were, which works but
is a poor substitute for the list simply being where you left it.

giancarloerra:
Fixes #6: Come back to where you were in a folder you have already been in.

## What this changes

Every folder you visit remembers the row you left it on, and coming back
lands there again. The memory is by the row itself, not its position: a
folder that was resorted or changed in between still comes back to the
same game, found by what it launches. When that exact row is gone, the
cursor lands where it stood, on whatever slid into its place.

- Kept per folder, keyed by system and place, up to 200 folders, oldest
  dropped first. Saved in `state.toml` with the resume state, so it
  rides along when a game takes the screen and starts fresh on a cold
  start, like the rest of where you were standing.
- Restoring after a game applies the saved cursor only when the whole
  saved walk could be replayed: a folder that vanished in between means
  the number belongs to a folder never reached, so the surviving parent
  re-lists on its own remembered row instead.
- Favouriting or hiding a game deep in a long list no longer snaps the
  cursor away: the folder is re-listed around the row you were on,
  including when the row you were on is the one that was hidden.
- The groups remember too: Console left on Neo Geo comes back to
  Neo Geo even after a walk through Computer, because each group writes
  down the system it was left on by id. The systems list is one shared
  list state, so its bare index meant a different machine in every
  group.

## Why

Opening a folder of thousands, walking to the game you care about, then
stepping out for a moment cost the whole walk again. The crumbs already
remembered an index for walking straight back out; an index goes stale
whenever rows are added, hidden or resorted, so this remembers identity
and falls back to position only when the row itself is gone.

## How it was checked

- [x] `cargo test`
- [x] `cargo fmt --check`
- [x] Run on a real MiSTer

## Licensing

- [x] I agree

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,9 @@
 # Binaries staged for the card; built, not committed.
 /deploy/Scripts/.config/degauss/degauss
 
-# Written by Degauss itself when settings change
+# Written by Degauss itself: options, and where you were
 /settings.toml
+/state.toml
 
 # Staged for the card at build time from assets/logos; not committed twice.
 /deploy/Scripts/.config/degauss/logos/
--- a/README.md
+++ b/README.md
@@ -96,7 +96,8 @@ Enter, Escape, Space and Tab.
 - **Nothing written to your card** except favourites you ask for, its own
   index, the `settings.toml` holding what you changed in Options, and a
   small `state.toml` remembering where you were: every folder you visit
-  keeps the row you left it on until the machine is powered off, found
+  keeps the row you left it on for the whole sitting, riding along when
+  a game takes the screen and starting fresh on a cold start, found
   again by the game itself rather than by its position, so a folder whose
   contents changed still comes back to the same game. An AmigaVision
   title also needs its `shared/ags_boot` file written, which is how those
--- a/README.md
+++ b/README.md
@@ -98,8 +98,9 @@ Enter, Escape, Space and Tab.
   small `state.toml` remembering where you were: every folder you visit
   keeps the row you left it on until the machine is powered off, found
   again by the game itself rather than by its position, so a folder whose
-  contents changed still comes back to the same game. AmigaVision titles
-  also need its own `shared/ags_boot` written to start the right game.
+  contents changed still comes back to the same game. An AmigaVision
+  title also needs its `shared/ags_boot` file written, which is how those
+  games are started.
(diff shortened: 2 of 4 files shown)


giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

Issue #7, opened 2026-08-27
Rebuild one system


giancarloerra:
### What are you trying to do

Adding games to one system means rebuilding every system. **Rebuild lists**
throws the whole cache away and reads the card again, a hundred and more
systems, when the only thing that changed was one folder.

On a large card that is a minute of waiting to pick up a handful of new
files, and it is the thing you do most often, because adding games to one
system at a time is how a card actually grows.

### What you have in mind

Two things, and the second one is nearly free.

**Rebuild the system you are standing in.** In the contextual menu, beside
the things that already act on the folder on screen: `Random game in this
folder`, `Search this folder`, `Hide this`, `Change view`. It needs no
picker, because the system you want is the one you are already in.

**A pass that only reads what changed.** `Rebuild lists` in Options gains a
value the way `View` and `Text` have one: `Everything`, as now, or
`Only what changed`.

Neither needs new machinery.

The cache is already one file per system, one hundred and fifteen of them on
a full card, plus an index holding the counts. `build_system()` builds one
and `save_system()` writes one.

The rebuild already works one system at a time: `build_one_system()` takes a
system off a list, reads it, says `Caching the card` with the name and how
far along it is, and yields between frames so the screen keeps moving. What
makes it rebuild everything is only that the list it works through is filled
with every system:

    let mut left: Vec<usize> = (0..total).collect();

Aiming it at one system is that vector holding one entry.

`Only what changed` exists too. `start_build()` already takes a `forced`
flag, and when it is false a system already written down is skipped rather
than read again. `Rebuild lists` passes `true`, so nothing uses the cheaper
path today.

**One thing that would go wrong if this is done naively.** `start_build()`
begins with a fresh, empty index and inserts one summary per system as it
goes. A single-system pass starting the same way would finish holding that
one system's counts and nothing else, quietly wiping the other hundred and
fourteen. A rebuild of one system has to fold its summary into the index
that is already there.

Done when:

- [ ] the contextual menu rebuilds the system on screen
- [ ] `Rebuild lists` offers `Everything` and `Only what changed`
- [ ] a single-system rebuild updates that system in the index and leaves the
      rest of it alone
- [ ] the counts shown after a single-system rebuild are right for every
      system, not only the rebuilt one
- [ ] the readme Settings table says what the two values do

Deliberately not a picker screen. A list of systems to choose from is the
browse screen you are already looking at, and there is no dropdown on a
stick and four buttons.

### What you tried instead

[...]

PR #17, opened 2026-08-29
Rebuild one system without reading the whole card

### LINKED ISSUE #14: Inside Favourites the context menu says Add to favourites, and Remove would do nothing
### Degauss version

0.2.0

### Display

Both

### What happened

Browsing the master Favourites from the root and pressing X on a game
offers **Add to favourites**. Everything in that folder is a favourite
already; the entry that makes sense there is **Remove from favourites**.

The label is only the visible half. Both directions resolve the selected
row the same wrong way, so both misbehave:

The context menu asks `favorites.holds(path)` with the path of the row
under the cursor (`open_context`, src/app.rs:3364-3368). Inside Favourites
that row is the favourite's own `.mgl` file, but `holds` keys favourites
by the **game path the `.mgl` points at** (`by_target`,
src/favorites.rs:86-88). The `.mgl`'s own path is never a key, so `holds`
answers false and the menu offers Add.

Choosing Add then goes down the self-describing-file path: an `.mgl` is
linked rather than described (`favorite_mgl` returns Ok(None) for it), so
Degauss would create a favourite pointing at a favourite.

And if the label were fixed alone, Remove would still do nothing:
`remove_favorite` looks the file up with `favorites.file_for(&game)`
(src/app.rs, remove_favorite), which is the same target-keyed lookup, so
it finds nothing for a path that is itself the favourite file.

The fix wants to be at the resolution, not the label: when the selected
row's path lies inside the favourites folder, it IS the favourite file.
Resolve it through `favorites::target_of` for the heart and the menu
question, and use the file itself for removal. Cores favourited as
symlinks need the same treatment.

### How to bring it back

1. Favourite any game from its own system; confirm the heart.
2. From the root open **Favourites**, then the folder it went to.
3. Press X on the game.

It offers **Add to favourites**. Expected: **Remove from favourites**,
and choosing it removes the `.mgl` under the cursor.

### The log

```text
Nothing is logged: no operation fails
```

### What the card looks like there

```text
Any card with at least one favourite. The `.mgl` files in
`_@Favorites/<folder>/` carry absolute `path="..."` attributes and resolve
fine; only the menu's lookup goes by the wrong key.
```

giancarloerra:
Part of #7: Rebuild one system. This implements the first half, the
contextual-menu rebuild of the system on screen.

Fixes #14: Inside Favourites the context menu says Add to favourites,
and Remove would do nothing. The second half, a
`Rebuild lists` value that only reads what changed, stays open on #7: what
"changed" means for a system spread over several folders needs its own
design pass, and half-answering it here would bake the wrong answer in.

## What this changes

The contextual menu gains **Rebuild this system** while a system is open,
beside the entries that already act on the folder on screen. It reads that
one system again from the card as it is now and rewrites only its cache
file and its line in the index; the other hundred-plus systems are not
read and their summaries are untouched.

- The system's folders are asked again at rebuild time, so a configured
  folder that appeared since startup is picked up. Which menu folder the
  core sits in stays a full-rebuild question.
- With every folder for the system gone, the cache is left as it was and
  a message says so: what to do about a vanished system is the full
  rebuild's decision.
- On a card that has never been indexed there is no index to fold into,
  and none is invented: a one-system index reaching the disk would make
  startup, which trusts an index that exists, bring up a card with one
  system on it. Pinned by test at the cache layer.
- A refresh that collides with a running full rebuild tells the build's
  index copy too, so the build finishing does not overwrite the refresh
  with what it saw earlier.

Device testing shaped the words and one adjacent repair:

- The refresh overlay says `Indexing <system> list`, not
  `Reading <system>`, which is the message for opening one; the full
  build's overlay says `Indexing all systems`, which is true on the
  first start too, where "rebuilding" described building something
  never built.
- The Options entry is `Rebuild all system lists`, so it and
  `Rebuild this system` read as the pair they are.
- The Favourites shelf now hearts every playable row, and its
  contextual menu works (#14): everywhere else a row is resolved by the
  game a favourite points at, but shelf rows ARE the favourite files,
  so the shelf showed no hearts, offered Add over things already there,
  and Remove found nothing to remove. The shelf now answers from what
  it is: every playable row a favourite, Remove removing the file under
  the cursor.

## Why

Adding games to one system meant rebuilding every system: a minute and
more of reading the whole card to pick up a handful of files in one
folder. The machinery was already per-system (one cache file per system,
built one at a time); what was missing was a way to aim it at one.

## How it was checked

- [x] `cargo test`
- [x] `cargo fmt --check`
- [x] Run on a real MiSTer

## Licensing

- [ ] I agree

## Summary by CodeRabbit

* **New Features**
  * Added **Rebuild this system** for refreshing an individual system’s list and cache.
  * Rebuilding all system lists now rediscovers newly added systems and cores.
  * Added progress updates and safer handling when systems are no longer available.
  * Favourites are now displayed and removed correctly while browsing the Favourites shelf.

* **Improvements**
  * Updated the rebuild option label to **Rebuild all system lists**.
  * Expanded documentation for rebuilding systems, installation, settings and follow-up steps.

* **Bug Fixes**
  * Prevented missing cache indexes from being created during repeated loading attempts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,8 @@ Enter, Escape, Space and Tab.
 - **The card is the truth.** Reorganise, rename or move files with any
   tool and the browser follows: nothing has to be re-imported or re-tagged.
   The index it keeps is only a copy of what the card already says, and
-  **Options -> Rebuild lists** brings it back in step.
+  **Options -> Rebuild all systems lists** brings it back in step,
+  and the contextual menu's **Rebuild this system** does one system alone.
 - **Full metadata.** Description, publisher, developer, release date,
   players and language, read from `gamelist.xml`.
 - **Favourites are MiSTer's favourites**, written into `_@Favorites` in
@@ -164,7 +165,7 @@ To remove Degauss, delete the `main=` line and the files above. Nothing
 else on the card is touched.
 
 The first run reads the card and writes an index, about a minute for a
-full one of 97k+ games. It never does that again on its own: **Options → Rebuild lists**
+full one of 97k+ games. It never does that again on its own: **Options → Rebuild all systems lists**
 is how you tell it the card has changed (for example after adding new games). New images and metadata are read on the fly.
 Adding games to one system does not need the whole card read again:
 **X → Rebuild this system** inside that system reads just its folders.
@@ -254,7 +255,7 @@ card.
 |---|---|
 | Scroll speed | How fast a held direction moves through the list |
 | Skip artwork faster than | Above this speed, pictures wait until the list stops |
-| Rebuild lists | Read the card again. Run it after adding games, cores or artwork |
+| Rebuild all systems lists | Read the whole card again. Run it after adding games, cores or artwork; for one system, the contextual menu's **Rebuild this system** is quicker |
 | View | Details, Tiled, List or Carousel |
 | Text | The typeface: Smooth, Pixel (a DOS font on whole pixels), and the bolder Smooth 2 and Pixel 2 |
 | Bottom bar while browsing | The strip with the time and the buttons. Menus always keep it |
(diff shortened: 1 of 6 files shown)


giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

PR #18, opened 2026-08-29
Choose what left and right do while browsing

### LINKED ISSUE #4: Option to change left/right behaviour
### What are you trying to do

Left and right change the scroll speed, 0.5x to 12x. In a folder of a few
hundred games that is right. In a folder of twelve thousand it is not: the
fastest speed still means holding a direction and watching names go past,
and the way to reach the S titles is the letter grid, which is a button, a
grid to cross and another button.

The speed is also something some people set once. The two most reachable
directions on the stick are permanently bound to it.

It costs something else, in Tiled view, and the code already says so. Moving
a whole row at a time would leave the covers either side of the selected one
with no key that reaches them, because left and right are taken. So up and
down move one tile at a time instead, walking left to right through the
grid. That is a compromise made to work around this binding, not the way a
grid wants to be driven.

### What you have in mind

A setting for what left and right do while browsing. Global, like the view
is, rather than per folder.

- **Speed**, exactly as now, and the default so nothing changes for anyone
  who does not go looking.
- **Letter**, jumping to the previous and next first letter in the list.
- **Page**, a screenful at a time.
- **Direction**, where they are simply left and right.

Letter and Page complement each other rather than competing. Letters are
uneven, and S in a large folder is enormous; a page is always a page.

Nothing here is new machinery. `first_letter()` already exists and the
letter grid already derives its letters from the list the same way.
`move_items()` already moves by any amount and already wraps at the ends the
way the rest of the list does, so Page is that call with the window size.

**Direction is the one that changes another control.** In List, Details and
Carousel the stride is one, so left and right land on the same neighbours up
and down do, and it is simply a second way to drive them for anyone who
thinks of a carousel as horizontal. Tiled is the case that matters: with
left and right free, up and down can go back to moving a whole row, which is
how a grid should work, and the compromise above goes away.

So on Direction, and only on Direction, up and down move by a row in Tiled.
On Speed, Letter and Page they keep moving one tile at a time, because
otherwise the covers beside the selected one would again have no key.

**The badge only makes sense for Speed.** The box that flashes the scroll
speed after a change should not appear at all on the other three: there is
nothing to report that the list itself does not already show.

Because this changes what the controls do, it has to be right everywhere the
controls are described:

- [ ] the setting, defaulting to Speed
- [ ] Letter, Page, Direction
- [ ] up and down move by a row in Tiled on Direction, and only there
- [ ] the badge appears on Speed only
- [ ] the bottom bar legend, which reads `<-> Speed` in its wide form
- [ ] the Help screen, which says "Stick left/right: change speed or setting"
- [ ] the readme control table, which says "scroll speed, 0.5x to 12x. In a
      menu, change the setting"
- [ ] the readme Settings table gains the row
- [ ] `degauss.toml`, so it is documented where the other defaults are
- [ ] left and right keep doing what they do on every screen that is not
      browsing: the letter grid, the options screens and the contextual menu
      each take their own branch already and none of them should change

### What you tried instead

Tried several iterations of controls mapping, and decided in the end is better to offer an option for users to customise the experience to their preference (that likely changes depending on items in folders).

giancarloerra:
Fixes #4: Option to change left/right behaviour.

## What this changes

A new **Left and right** row in Options decides what the two buttons do
while browsing games: **speed** (scroll speed, exactly as before, the
default), **letter** (previous or next first letter in the folder),
**page** (a screenful at a time), or **direction** (move the selection
itself; in the Tiled view this frees up and down to move a whole row).
Live from Options, kept in `settings.toml`.

- Letter and Page repeat while the key is held, at the same cadence a
  held up or down uses; only the speed ladder stays one step per press,
  where a held repeat would run the whole ladder off one touch.
- The row sits below Skip artwork faster than, and its default value
  shows as Scroll speed change; the token settings.toml stores is
  unchanged, so a written-down choice keeps meaning the same thing.
- PgUp and PgDn now page on every screen, not only while browsing games.
  The Find grid moves by one cell instead, its rows not being pages.
  Before this they did nothing off the games list.
- A press that dismisses a message is consumed by the dismissal and does
  not also jump: dismissing must never move the list underneath the
  message.
- `degauss.toml` documents the key commented out on purpose: versions
  from before this key refuse settings they do not know, so an active
  line would stop an older build from starting after a downgrade. The
  Options screen is the way to set it.

## Why

Left and right were fixed on scroll speed. On a folder of thousands the
speed is set once and the two best-placed buttons on the pad then do
nothing; a letter jump or a page at a time is what that walk actually
needs, and which one is a matter of taste, so it is a setting.

## How it was checked

- [x] `cargo test`
- [x] `cargo fmt --check`
- [x] Run on a real MiSTer

## Licensing

- [x] I agree

## Summary by CodeRabbit

- **New Features**
  - Added configurable left/right browsing modes: speed, letter jumps, page jumps, and directional movement.
  - Added a **Left and right** setting in Options, with the selected behaviour shown in help and navigation hints.
  - Added dedicated Page Up and Page Down actions.
  - Directional mode supports horizontal movement and row-by-row navigation in tiled views.

- **Bug Fixes**
  - Preserved speed behaviour as the default when no preference is configured.
  - Improved horizontal key-repeat handling for each browsing mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Source available, written in Rust and using Slint, and licensed for non-commerci
 | Control | Does |
 |---|---|
 | **up / down** | move through lists |
-| **left / right** | scroll speed, 0.5x to 12x. In a menu, change the setting |
+| **left / right** | scroll speed, 0.5x to 12x, or what the **Left and right** setting says: letter jumps, page jumps, or plain movement. In a menu, change the setting |
 | **A** (enter) | open a folder, launch a game |
 | **B** (escape) | back, out of the folder |
 | **X** (tab) | context menu: random game, random favourite, keep or drop a favourite, jump to letter, search, hide a row, rebuild this system, change view, etc. |
@@ -254,6 +254,7 @@ card.
 | Setting | Does |
 |---|---|
 | Scroll speed | How fast a held direction moves through the list |
+| Left and right | What left and right do while browsing: Speed (the default), Letter, Page or Direction. Direction frees up and down to move a whole row in Tiled |
 | Skip artwork faster than | Above this speed, pictures wait until the list stops |
 | Rebuild all system lists | Read the whole card again. Run it after adding games, cores or artwork; for one system, the contextual menu's **Rebuild this system** is quicker |
 | View | Details, Tiled, List or Carousel |
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ card.
 | Setting | Does |
 |---|---|
 | Scroll speed | How fast a held direction moves through the list |
-| Left and right | What left and right do while browsing: Speed (the default), Letter, Page or Direction. Direction frees up and down to move a whole row in Tiled |
+| Left and right | What left and right do while browsing: Scroll speed change (the default), Letter, Page or Direction. Letter and Page repeat while held; Direction frees up and down to move a whole row in Tiled |
 | Skip artwork faster than | Above this speed, pictures wait until the list stops |
 | Rebuild all system lists | Read the whole card again. Run it after adding games, cores or artwork; for one system, the contextual menu's **Rebuild this system** is quicker |
 | View | Details, Tiled, List or Carousel |
--- a/degauss.toml
+++ b/degauss.toml
@@ -41,6 +41,19 @@ art_limit = 3
 # Which view to open in: details, tiled, list or carousel.
 layout = "details"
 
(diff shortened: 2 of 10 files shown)


giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

PR #19, opened 2026-08-29
Switch palettes with themes kept as files

### LINKED ISSUE #3: Theming with options and named themes
### What are you trying to do

Every colour Degauss draws with is already a setting. `degauss.toml` has a
`[colors]` block naming ten roles.

Two things stop it being useful beyond one person editing one file:

There is nowhere to keep more than one. Trying an amber scheme means
overwriting the colours you had, and the only way back is remembering what
they were.

The wordmark does not follow. It is drawn in the three brand colours, so a
palette those clash with looks wrong no matter how good the rest of it is,
and there is no way to say "not those three, here".

### What you have in mind

**Themes as files.** A `themes/` folder beside `degauss.toml`, one `.toml`
per theme, each holding the same `[colors]` block the main file already
takes, with the filename as its name. A `Theme` row in Options cycles
through whatever is in the folder. Ship three or four to start from and to
copy: the current palette, amber, green, and something monochrome.

Reading them needs nothing new. The colour table already parses from TOML,
already validates `#rrggbb` at load, and already fills in per-field defaults,
so a theme file that names three colours and leaves the rest alone works.

**The wordmark takes a colour from the theme.** The renderer can already
recolour an image, and the wordmark is built for it: 600x150 with the shape
in the alpha channel, 44% transparent, 42% opaque, 13% antialiased edge. Set
a colour and it becomes a flat silhouette with clean edges; leave it unset
and it draws in its own colours.

So one more key in the theme, something like `logo`, which is either absent
for the original colours or a colour for a flat one. Black and white are
what people will reach for, but there is no reason to limit it to those.

Two places draw the wordmark, the About screen and the brand at the bottom
of the group list, so it applies in both.

**What this deliberately does not include: editing colours inside Degauss.**
Choosing a colour with a stick and four buttons is a poor interface for a
job that a text editor does well, and the card is already in your hand. Read
themes, do not write them.

Done when:

- [ ] `bar` documented in `degauss.toml`, which has the other nine and not
      that one, so nobody knows it can be set
- [ ] the palette documented in the readme, since it already works
- [ ] `themes/` read at startup, a theme applied by name
- [ ] `Theme` in Options, cycling what the folder holds
- [ ] a theme change repaints the whole screen, not only what was dirty
- [ ] the wordmark takes its colour from the theme, unset meaning original
- [ ] three or four themes shipped
- [ ] a theme naming only some colours works, the rest falling back

### What you tried instead

Editing `[colors]` in `degauss.toml` directly, which works and is how the
current palette was arrived at. It is keeping more than one, and getting the
wordmark out of the way of it, that there is no answer for.

Once the install folder moves to `Scripts/.config/degauss`, `themes/` should
land there.

giancarloerra:
Fixes #3: Theming with options and named themes.

## What this changes

A **Theme** row in Options switches between named palettes, and a theme
is one `.toml` file in a `themes/` folder beside `degauss.toml`, its file
stem being its name. Three ship as starting points: `Amber` after the
amber phosphor, `Mono`, the standard greys with the colour taken out,
and `Blue-Orange`, slate blue with an orange highlight. None of them
names `favorite`, so the hearts keep the user's own favourite colour
under every palette: red, unless it was changed in `[colors]`. The built-in state shows as `Standard`, and the remembered
name is looked up without regard to case, which is all the card's
filesystem could tell apart anyway.

- A theme names any subset of the ten colour roles, as bare keys or under
  a `[colors]` header, so the block from `degauss.toml` pastes in
  unchanged. A role set both ways is refused by name; a typo'd role is
  refused with the key, exactly as in `degauss.toml`.
- A theme overlays the user's configured `[colors]`: roles it does not
  name show through from `degauss.toml`, never from whichever theme was
  on before. One more key, `logo`, draws the wordmark as a flat
  silhouette; without it the wordmark keeps its own colours.
- Switching repaints everything live, covers included. **Standard** is
  the palette from `degauss.toml`, and the choice is remembered by name
  in `settings.toml`; if the file it names is missing or broken at the
  next start, the standard palette is used and a message says so.
- A broken theme file never stops Degauss starting: it is reported on
  screen and the rest still load. Two names differing only by case are
  both refused, because the card's filesystem cannot tell them apart.
- The three shipped files are updater-managed; the readme says to copy
  one under a new name to make it your own.

## Why

Every colour was already a role in `[colors]`, but trying a palette meant
editing the one config by hand and there was no way to keep more than one
or to switch at the machine. Files keep the editing where a text editor
is good at it; the Options row does the switching where the machine is.

## How it was checked

- [x] `cargo test`
- [x] `cargo fmt --check`
- [x] Run on a real MiSTer

## Licensing

- [x] I agree


giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

PR #20, opened 2026-08-29
Find games where MiSTer finds them: USB, network, CIFS, card

giancarloerra:
Closes no issue: this came out of reviewing what a card with games on a
USB stick sees, which was nothing.

## What this changes

Systems on external storage are discovered without any setup. The
default `game_roots` now walk the exact order MiSTer's own loader
searches (`findPrefixDir` in Main's `file_io.cpp`): `/media/usb0..5/
games`, `/media/network/games`, `/media/fat/cifs/games`,
`/media/fat/games`, and `/media/fat` for Arcade.

And the promise the configuration always made is now kept: for each
folder name the FIRST root that has it wins, instead of stitching one
listing from every root that matched. A system on both the stick and
the card browses from the stick, exactly as the stock menu would load
it. Folder aliases still resolve independently, each on its own root:
MegaDrive on the stick and Genesis on the card still browse as one
system.

- Unmounted roots cost one failed stat each and are skipped; an SD-only
  card resolves exactly as before, pinned by test.
- Behaviour change for a card keeping the same folder in two places: it
  used to show both stitched together, it now shows the higher-priority
  one, which is what MiSTer itself loads.
- Storage is found when Degauss starts: plug in first, or restart. A
  system moved between storages changes its paths, so its listing is
  stale until Rebuild this system or a full rebuild. The readme says
  both, in an External storage section.
- A layout the defaults do not cover remains one `game_roots` edit away;
  the format and the user's ordering are untouched.

## Why

MiSTer supports games on USB sticks, the network share and the CIFS
mount, and the stock menu finds them with zero configuration. Degauss
looked only at the card unless `game_roots` was edited by hand, which
made it the one part of the setup that did not follow the card's own
rules.

Five tests pin the semantics: USB priority, card fallback, aliases
across roots, unmounted roots ignored, a folder reached through a
symlink.

## How it was checked

- [x] `cargo test`
- [x] `cargo fmt --check`
- [x] `./scripts/rehearse-migration.sh`
- [x] Run on a real MiSTer

## Licensing

- [x] I agree

## Summary by CodeRabbit

* **New Features**
  * Games can now be discovered automatically across USB storage, network shares, CIFS mounts and the console’s internal storage.
  * Storage locations are checked in a defined priority order, with fallback support when locations are unavailable.
  * Different game systems can resolve to different storage locations, including supported aliases.

* **Bug Fixes**
  * Unmounted storage locations are now ignored during game scanning, improving reliability.

* **Documentation**
  * Added guidance on external storage discovery, search precedence, startup timing, relocated games and custom storage configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,22 @@ Its source is at
 To remove Degauss, delete the `main=` line and the files above. Nothing
 else on the card is touched.
 
+### External storage
+
+Games on a USB stick, the network share or the CIFS mount are found
+without any setup, in the order MiSTer's own loader searches:
+`/media/usb0` to `usb5`, then `/media/network`, then `/media/fat/cifs`,
+then the card, each under its `games` folder. For each system folder the
+first place that has it wins, so a system kept on both the stick and the
+card browses from the stick, exactly as the stock menu would load it.
+Different folders of one system may resolve in different places.
+
+Two things to know. Storage is looked for when Degauss starts, so plug
+the stick in first (or restart after); and moving a system between
+storages changes its paths, so its listing is stale until **Rebuild this
+system** or a full rebuild. A layout the defaults do not cover is one
+`game_roots` edit in `degauss.toml` away.
+
 The first run reads the card and writes an index, about a minute for a
 full one of 97k+ games. It never does that again on its own: **Options → Rebuild all system lists**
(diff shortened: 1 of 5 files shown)


giancarloerra:
@coderabbitai review

PR #64, opened 2026-09-11
Skip malformed archives and unsupported ZIP members while indexing

### LINKED ISSUE #49: Handle malformed and unsupported ZIP content without failing an entire system index
## Problem

Degauss currently treats every ZIP validation problem as an error for the complete archive. That error propagates out of the archive browser and aborts the system-cache build, so one malformed archive or one unsupported member can prevent healthy content elsewhere in the same system from being published.

The same path currently conflates three different cases:

1. a structurally malformed or unreadable ZIP;
2. a valid ZIP containing an individual member that Degauss or MiSTer Main cannot use;
3. a valid legacy ZIP filename that is not exact UTF-8.

Relevant implementation:

- [A nested ZIP member currently invalidates the complete outer archive](https://github.com/giancarloerra/Degauss/blob/4e65edea4b03de1bcb658d606f218b7a0404604f/src/zip.rs#L295-L345)
- [A non-UTF-8 member name currently invalidates the complete archive before its flags and extra fields are considered](https://github.com/giancarloerra/Degauss/blob/4e65edea4b03de1bcb658d606f218b7a0404604f/src/zip.rs#L420-L448)
- [The archive error propagates from browsing](https://github.com/giancarloerra/Degauss/blob/4e65edea4b03de1bcb658d606f218b7a0404604f/src/browse.rs#L622-L635)
- [The archive error currently aborts the system-cache traversal](https://github.com/giancarloerra/Degauss/blob/4e65edea4b03de1bcb658d606f218b7a0404604f/src/cache.rs#L318-L364)
- [The existing regression test preserves the previous complete cache by failing the whole replacement](https://github.com/giancarloerra/Degauss/blob/4e65edea4b03de1bcb658d606f218b7a0404604f/src/index_job.rs#L495-L510)

## MiSTer compatibility boundary

Degauss must only expose archive members that the installed MiSTer Main can address and launch byte-exactly.

- [MiSTer Main splits a launch path at the first `.zip`](https://github.com/giancarloerra/Degauss-Main/blob/0651979d53f570c29f8772e124ae603297830954/file_io.cpp#L198-L212), and [explicitly rejects a second `.zip` in the path](https://github.com/giancarloerra/Degauss-Main/blob/0651979d53f570c29f8772e124ae603297830954/file_io.cpp#L1548-L1552). Degauss therefore must not present an inner ZIP as launchable. It should skip that member while retaining supported sibling members.
- A legacy filename must not be converted lossily. Degauss should inspect the ZIP UTF-8 flag and Unicode Path extra field and use a decoded name only when the exact path emitted by Degauss is proven to resolve through current MiSTer Main. If it cannot round-trip exactly, skip that member with an accurate unsupported-filename warning.
- Extraction is not an acceptable generic fallback. It adds storage, startup, cancellation, and very large disc-image costs, and would change Degauss's current central-directory-only design.

The objective is deterministic, resilient handling of every ZIP condition. It is not to display an entry that MiSTer Main cannot launch.

### 1. Separate archive-level and member-level failures

- Treat an unreadable, truncated, structurally inconsistent, or otherwise invalid central directory as an archive-level failure.
- Skip that archive subtree, record a warning, and continue indexing healthy siblings and the rest of the system.
- Treat a condition confined to one member as a member-level failure. Skip only that member, record a warning, and retain supported members from the same archive.
- If ambiguity involves a group of entries, such as duplicate or ASCII-case-colliding paths, skip the complete conflicting group rather than choosing one silently.
- Never publish an incomplete row for a rejected archive or member.

### 2. Handle every current validation class explicitly

- Nested ZIP member: skip the inner archive member because current MiSTer Main cannot launch it; retain supported outer members.
- Filename encoding:
  - honour the ZIP UTF-8 flag;
  - validate and use the Info-ZIP Unicode Path extra field when present;
  - accept only a representation proven to round-trip byte-exactly through Degauss's cache, generated MGL, and current MiSTer Main;
  - otherwise skip only that member and report `unsupported legacy ZIP filename encoding`, not `malformed ZIP`.
- Encrypted or compressed-patch member: skip the member with its exact reason.
- Unsupported compression method: skip the member with its method number.
- Unsafe or non-round-trippable name, traversal segment, separator, control character, filename limit, or complete native-path limit: skip the member with its exact reason.
- Duplicate or case-ambiguous paths: skip the conflicting entries and retain unrelated entries.
- Invalid ZIP64 metadata, multi-disk structure, central-directory bounds/signature/count/offset inconsistency, or truncation: skip the archive as structurally invalid.
- Keep existing support for valid stored and deflated entries and valid ZIP64 archives unchanged.

### 3. Preserve the rest of the index transaction

- A rejected archive or member must not stop healthy folders, files, or other ZIPs in that system from being indexed and published.
- A system containing only rejected content may complete with zero usable games and a warning; it must not appear successful without the warning.
- Cancellation must preserve the previous complete cache exactly as it does now.
- A genuine system-root read failure or cache-publication failure must remain a failed transaction and preserve the previous complete cache. Do not turn those failures into an empty successful index.
- Gamelist and Artwork Pack source preparation must use the same retained healthy game tree and must not replace the original ZIP cause with a secondary `prepared cache is missing` message.

### 4. Make warnings useful without overwhelming the UI

- Finish the operation as `Finished With Problems` when any archive or member was skipped.
- Show a concise, deduplicated summary identifying the system, archive, and reason.
- Keep every complete diagnostic, including the member path where safely representable, in `/tmp/degauss.log` for the current Degauss run.
- Preserve the first/root cause when a later source-preparation step cannot proceed because of it.
- Keep warnings actionable and do not label a standards-compliant but unsupported legacy encoding as corruption.

### 5. Preserve performance and responsiveness

- Continue reading only ZIP structure and central-directory data while indexing. Do not extract or decompress game payloads.
- Do not add another complete library traversal or extra work to ordinary non-ZIP folders.
- Keep progress, input, and cancellation responsive during large-library indexing.
- Preserve Degauss's exit-on-launch architecture and do not add a background service.

## Acceptance checks

- A system containing healthy ordinary games plus one structurally corrupt ZIP publishes the healthy games and reports the skipped archive.
- A valid outer ZIP containing supported games plus an inner ZIP publishes the supported outer members, omits the inner ZIP, and reports it.
- A valid archive with a legacy CP437 name, a valid UTF-8-flagged name, a valid Unicode Path extra field, and an invalid UTF-8-flagged name exercises each encoding path without lossy conversion.
- Encrypted, compressed-patch, unsupported-compression, unsafe-name, overlong-name, duplicate-path, and ASCII-case-collision fixtures skip only the affected member or conflict group.
- Truncated central directory, invalid signature/bounds/count/offset, invalid ZIP64, and multi-disk fixtures skip the complete affected archive while the system continues.
- A rebuild with a previous complete cache publishes the newly indexed healthy subset only after the system transaction succeeds; cancellation and genuine root/publication failures retain the previous cache.
- Both Gamelist and Artwork Pack modes complete with the same healthy game set and retain the original warning cause.
- The completion UI is responsive, reports `Finished With Problems`, and the current-run log contains the full diagnostics.
- A retained supported ZIP member is launched successfully through the production Degauss-to-MiSTer-Main path on a real MiSTer. Rejected members never appear as launchable rows.
- A representative full-library MiSTer benchmark shows no material regression for ordinary folders or supported ZIP libraries.

## Backward compatibility

- No setting, cache-format migration, reset, or forced rebuild is required solely by this change.
- Existing ordinary folders and supported ZIPs retain their current names, artwork, metadata, hierarchy, and launch behaviour.
- The previous complete cache remains recoverable until a replacement system cache has been built and published successfully.

giancarloerra:
## Summary

- A ZIP whose central directory does not hold together (truncated, unreadable or inconsistent directory; invalid signature, count, size or offset; invalid ZIP64 metadata; multi-disk structure; a record whose sizes or local-header offset do not fit the archive; a masked local header; an archive path holding an earlier `.zip`) is skipped whole with its path and reason. Nothing from it is listed and the healthy folders, files and other archives of the same system are still indexed and published in the same transaction.
- A condition confined to one member skips only that member, with its exact reason, and the supported siblings stay: an inner `.zip` member (Main refuses a second `.zip` in a target), an encrypted entry, a compressed-patch entry, another compression method (named by its number), a traversal or empty path segment, a backslash or colon, a control character, leading or trailing whitespace, a member filename over 260 bytes or a native target over 1,023 bytes.
- Duplicate or ASCII-case-ambiguous paths, including a file whose name is also a directory, skip the whole conflicting group rather than choosing one entry, as `duplicate or case-ambiguous member path`, with the group's key after the reason in the log and report lines so the summary counts every conflict member of an archive on one line. A member already skipped for its own reason still counts as part of a group, because Main's case-insensitive lookup can land on its record.
- Member names are used exactly as their raw central-directory bytes, because Main locates a member by comparing the requested bytes with those raw bytes without decoding them and without reading the UTF-8 flag or the Unicode Path extra field. A raw name that is valid UTF-8 is accepted whether or not the archive sets the UTF-8 flag. A name that is not valid UTF-8 is skipped as `unsupported legacy ZIP filename encoding` whether or not the archive flags it UTF-8; a wrong flag is noted after the reason (`flagged UTF-8`) in the log and report lines rather than given a second reason. The Unicode Path extra field is validated (version 1, matching CRC-32 of the raw name) and its decoded name is written after the reason in the log and report lines to identify the member; it is never used as a launch name and no name is decoded lossily.
- Every central-directory record is checked for the archive-level conditions before any member can be skipped on its own, matching the order in which Main refuses an archive at open. An encrypted stored member fails the archive only in the case where Main's stored-size check fails it (method and DOS time word zero) and is otherwise skipped on its own.
- Beyond the issue's archive-level and member-level classes: a folder inside an archive that would sit deeper than the folder depth the index walks (twelve levels below the system start, the folders holding the archive and the archive included) is skipped with everything under it, as `<archive>/<folder>: skipped: folder is past the maximum depth of 12 below the system start`; the archive's other members stay, and `--report` and `--audit` name the same folder with the same reason. Before this change any place past that depth, a folder inside an archive included, failed the whole system. A folder on the card that deep remains the system's failure.
- A rebuild that skipped anything finishes as **Finished With Problems** with one deduplicated line per system, archive and reason (member skips are counted per archive and reason). The index and the audit (`--report`, `--audit`) write every skipped member to `/tmp/degauss.log` as `zip <archive>: member <name>: <reason>`, once per archive each time one of them reads it, a skipped archive as `zip <archive>: skipped: <error>` (the audit writes the same line when one of an archive's folders can no longer be listed because the archive changed after its root listing) and a skipped folder as `zip <archive>/<folder>: skipped: <reason>` at the moment they skip it, so the log keeps the full diagnostic even when the summary is dropped later or the report has room for only the first few lines. No other read of an archive logs: the on-demand listing of a system that is not indexed, the Details view, the Artwork Pack passes, the launch check and favourite resolution read archives without repeating the block.
- Gamelist and Artwork Pack source preparation use the same retained healthy game tree and carry the scan warnings instead of failing the group with a secondary `prepared cache is missing` message. Beyond the issue's scope, so that the summary is on screen: the completion messages of Rebuild This System List on a Pack-prepared system, of the preparation on opening such a system and of a change of its game data source now list every warning in full (the scan, storage and install warnings included) where they previously said `Saved with a storage warning; see degauss.log`, and the publication warnings of a single-system rebuild are prefixed with the system name like a full build's. A full build logs each of these warnings once, where the recovery step and the terminal each logged it before; a Pack group whose game data source is unresolved is left out of the build and prepared by the recovery step after the build has finished, and that step logs its warnings itself.
- The scraper's Refreshing Lists step reports a skipped archive or member as the dashboard's last problem, prefixed with the system name and written to the log, without counting the system as a failed refresh: the system's list was replaced. Before this change the refresh path handed publication warnings to the failure branch, which labelled the refresh `Library refresh failed` and counted a system error; the scan warnings this change adds would have reached the same branch.
- Launch confirmation reports the skip reason for a member the archive holds but the reader skipped (a favourite or gamelist entry can still name one) instead of reporting it as missing or renamed.
- `--report` lists a skipped archive or member as an `unreadable` line with its reason and `--audit` counts it as a problem; the skipped member is named as `<archive>/<member>` built as text so a name starting with a slash cannot replace the archive path.
- A system holding only rejected content completes with zero games and the warning. A system root that cannot be read, a cancellation and a cache publication failure remain failed transactions that preserve the previous complete cache.

Fixes #49

## Compatibility

- No setting, cache-format migration, reset or forced rebuild is needed. Existing ordinary folders and supported ZIPs keep their names, artwork, metadata, hierarchy and launch behaviour; valid stored and deflated entries and valid ZIP64 archives are read as before, and the index still reads only the central directory.
- A single supported member left after skips keeps the legacy archive-level metadata binding; skipped members do not exist for Degauss.
- An archive whose central-directory records carry the ZIP64 disk sentinel (0xFFFF) was listed before and is now skipped whole; it was never launchable, because Main refuses that record when it opens the archive. Degauss keeps refusing a record that says disk 1 in a disk-0 archive, where Main accepts it, which only ever leaves an archive out.
- An empty archive at a path holding an earlier `.zip` was accepted before and now fails as an archive, since no member of such an archive is addressable by Main.
- The previous complete cache stays recoverable until a replacement system cache has been built and published.
- `scripts/test-main-zip.py` keeps compiling `src/zip.rs` on its own, unchanged: the reader does not log, so the standalone compile needs no hook.

## Tests

- `src/zip.rs`: a truncated, malformed-length or invalid-signature directory, a multi-disk end record or member disk number, a masked local header, an inconsistent record beside a member that would be skipped, and an ambiguous outer path each fail the archive; an encrypted stored member fails the archive exactly when Main refuses it; unsupported methods and flags skip only that member; a legacy CP437 name, a valid UTF-8-flagged name, an unflagged valid UTF-8 name, a valid Unicode Path field and an invalid UTF-8-flagged name are each listed or skipped byte-exactly without lossy decoding; a retained member cannot share a lowercase name or directory prefix with a skipped one; a control character in a skipped name is escaped only where it is printed; a skipped member is reported with its reason at launch confirmation instead of looking renamed.
- `src/cache.rs`: a corrupt archive is skipped and its siblings are published for each structural class (truncated, garbage, invalid signature, inconsistent count and offset, invalid ZIP64 end record, multi-disk end record, masked local header, ZIP64 sentinel member disk); an inner archive member is skipped and the outer members are kept; several rejected archives in one folder each lose only their own row, with healthy archives and a file between them kept; a folder past the depth limit inside an archive is skipped with everything under it while the archive's shallow member is published and a folder on the card that deep still fails the system; a skipped archive takes its progress counts and member lines back with it; a nested malformed archive is reported as a warning up to publication; the log holds the exact member line and the archive line with its complete error, and the launch check does not repeat them.
- `src/index_job.rs`: a corrupt ZIP is skipped and the healthy subset is published only after the system transaction succeeds; a root read failure keeps the previous cache and index.
- `src/source_cache.rs`: Gamelist and Artwork Pack modes stage the same healthy tree and carry the original ZIP cause.
- `src/browse.rs`: skipped archive members are audited, named as `<archive>/<member>` with the reason and its detail, and never listed as rows; a folder past the depth limit inside an archive is named by the audit with the same reason the index uses, once per folder, and the shallow member is counted; the audit writes the member line, the folder line and the archive line with its complete error to the log while a plain listing of the same archives writes nothing; an archive rewritten without one of its folders between its root listing and that folder's turn is named by the audit with the folder's error and logged as the archive.
- `src/ui_acceptance_tests.rs`: a rebuild with a corrupt archive and an inner archive member finishes as **Finished With Problems** naming the system, archive and reason, publishes the healthy games, and neither rejected item is reachable as a row; the Pack-prepared rebuild and the game data source switch show the archive and reason lines in their completion message, the full build of a Pack-prepared system logs its warning once, the single-system rebuild logs it once from the recovery step, and so does the recovery of a group whose unresolved game data source left it for the step after the build; a scraper list refresh of a system holding an archive with an inner archive member finishes without a system error, names the system, archive and reason as the last problem and in the log, publishes the retained member and drops the inner archive from the rows.
- `scripts/test-main-zip.py`: the encrypted and unsupported-method fixtures list the supported sibling, accept it for launch and refuse the mutated member for launch while Main's extract iterator refuses that member (the archive opens); a masked local header and a ZIP64 sentinel member disk are refused whole by both readers.

## Validation

- Host gates (`cargo fmt --check`, `cargo test`, the migration rehearsal, the RetroAchievements release check and `cargo clippy --target armv7-unknown-linux-musleabihf --all-targets -- -D warnings`) passed on the pushed head.
- `scripts/test-main-zip.py` was run against the pinned Main `lib/miniz` sources with verified hashes.
- Further review-round commits may follow on this branch.
- Device (headless, a synthetic fixture library on a MiSTer): `--list-systems`, `--report`, `--audit` and `--dry-run-launch` list the supported members, report the truncated archive, the inner archive member, the legacy-encoded member, the encrypted member and the case-ambiguous group with their reasons, generate the MGL for a retained member, and append the skip lines to `/tmp/degauss.log`; the same fixture on the previous release reports every affected archive whole as malformed. `--report` on a ZIP-holding system and `--audit` over a full library show no material change against the previous release.
- Still pending on hardware: the real launch of a retained ZIP member through the production Degauss-to-Main path, the interactive full-library index with the completion UI, and items 6 to 8 of `docs/testing/library-compatibility-hardware.md`.

## Summary by CodeRabbit

- **New Features**
  - ZIP archives retain healthy content while reporting skipped or unsupported members with specific reasons.
  - Archive audits and reports include skipped-member paths and explanations.
  - Rebuilds, source switches, and scraper refreshes surface detailed warnings, including affected systems and archives.

- **Bug Fixes**
  - Healthy content remains available when individual archive members fail.
  - Previous cache data is preserved when a system cannot be read safely.
  - Excessively deep archive folders are skipped without affecting valid members.
  - Concurrent log output remains intact.

- **Documentation**
  - Updated ZIP, CLI, and testing documentation to describe recovery, warnings, reporting, and validation behaviour.


giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

PR #77, opened 2026-09-14
Improve Theme Editor held colour adjustment and saving

### LINKED ISSUE #76: Improve Theme Editor held colour adjustment and saving
## Problem

The Theme Editor has two unnecessary points of friction:

1. In the visual RGB colour picker, left and right change the selected channel by one step but do not repeat when held. Large adjustments therefore require repeated taps. The exact hex editor already repeats up and down when held.
2. An existing theme created by the Theme Editor cannot be updated in place. The editor only offers `Save As`, and an existing name is rejected, so every edit requires creating another theme.

## Required behaviour

- In the visual RGB colour picker, holding left or right must continuously adjust the selected channel after the normal repeat delay.
- Releasing the direction must stop the adjustment immediately.
- Held left and right must continue to behave as they do now everywhere else in the Theme Editor.
- When the starting point is a theme previously created by the Theme Editor, offer both:
  - `Save Changes`, which updates that theme under its existing name.
  - `Save As`, which creates a separate theme under a new name.
- Built-in themes and theme files not created by the Theme Editor must remain protected from replacement and continue to offer `Save As` only.
- `Save As` must continue to reject an existing name.
- A failed `Save Changes` must leave the existing theme usable and show the real error.
- Saving must keep the updated theme selected and survive restarting Degauss.
- Existing custom themes, theme selection, cancellation, discard and deletion behaviour must remain compatible.

## Acceptance checks

- Hold left and right in each RGB channel and verify continuous adjustment, correct limits and immediate stopping on release.
- Confirm that exact hex editing still repeats up and down as before.
- Confirm that held directions do not start changing sources, fonts, names or confirmation choices unexpectedly.
- Edit a Theme Editor-created theme, use `Save Changes`, restart Degauss and verify that the same theme name contains the new values.
- Use `Save As` and verify that the original remains unchanged and name collisions are still rejected.
- Confirm that built-in and externally supplied themes cannot be overwritten.
- Exercise both save paths on the MiSTer's actual filesystem.

giancarloerra:
## Summary

- repeat held left and right only in the continuous RGB picker
- let Theme Editor-created themes be updated transactionally under their existing name
- keep built-in and externally supplied themes protected and preserve Save As behaviour

Closes #76

## Summary by CodeRabbit

- **New Features**
  - Save changes directly to existing editor-created themes.
  - Use held left/right input to repeatedly adjust RGB values.
  - Theme updates and settings saves now protect existing files if saving fails, restoring the previous version when necessary.

- **Bug Fixes**
  - Improved theme editor selection handling when available rows change.
  - Built-in and hand-written themes remain available through “Save as” only.

- **Documentation**
  - Updated theme editor guidance for saving, RGB adjustment, and safe theme replacement.


giancarloerra:
Superseded by and included in #82, which has now merged into main.

PR #79, opened 2026-09-14
Add configurable hold shortcuts for common browsing actions

### LINKED ISSUE #78: Add configurable hold shortcuts for common browsing actions
### Why expand the original request

The original request proposed using Select to cycle view modes. Degauss can also be launched from Scripts under the standard MiSTer Main binary, where Start and Select are not reliably delivered to the framebuffer application. Using either button would require changes outside Degauss and would not work consistently across supported launch methods.

A one-second hold on A, B, X or Y uses buttons that Degauss already receives in every launch method. It also replaces the two separate Hold X and Hold Y settings with one coherent shortcut system instead of adding another special case.

### Required Options interface

Add a **Shortcuts** section to General Options with four rows:

- Hold A
- Hold B
- Hold X
- Hold Y

Each row can independently select one of these actions:

- None
- Cycle View
- Random Game
- Random Favourite
- Add/Remove Favourite
- Game Information
- Search This Folder
- Jump to Letter

Remove the existing **Hold X (1s) to Add/Remove Fav** and **Hold Y (1s) for Random Game** rows from the visible Options interface. Their stored settings remain supported for compatibility as described below.

Start and Select are deliberately not included because their availability would depend on the MiSTer Main binary used to start Degauss. Short-press button assignments are not changed.

### Hold behaviour

- The hold threshold remains fixed at one second.
- Releasing before one second performs the button's normal short action.
- Reaching one second performs the configured shortcut once.
- Releasing after the shortcut fires must not also perform the short action.
- Hold shortcuts do not repeat.
- Another input, selection change or screen change cancels a pending hold so it cannot act on a different item or place.
- The gesture is armed only when its configured action is available in the current context. If the action is unavailable, the button behaves normally and is not delayed by the shortcut system.
- Hold shortcuts operate only from the browsing interface. Dialogues, Options, Actions, the scraper and other screens keep their normal controls.

### Context-dependent actions

Use the same implementations, scopes and availability checks as the corresponding existing Actions commands. Do not create separate shortcut-specific behaviour.

- **Cycle View:** available at every browsing place. Cycle through the existing views in their established order, save the result for that exact category, system or folder, and apply it immediately. This is the expanded form of the original issue request.
- **Random Game:** available while browsing games inside a system or folder. Use the current folder scope and the existing Random Game Behaviour setting.
- **Random Favourite:** available while browsing games. Use the same scope and Random Game Behaviour as the existing Random Favourite action.
- **Add/Remove Favourite:** available when a playable game is selected outside the master Favourites collection. Use the existing add-folder choice and removal behaviour.
- **Game Information:** available when a playable game is selected.
- **Search This Folder:** available only where the existing Search This Folder action is offered.
- **Jump to Letter:** available only where the existing Jump to Letter action is offered.
- **None:** disables the hold shortcut for that button.

An unavailable shortcut must not produce a misleading operation, change scope, or consume the normal button action.

### Backward compatibility

Existing settings must work without a reset, rebuild or manual migration.

- If the new Hold X binding is absent and the legacy `hold_x_favorite` value is true, resolve it as **Hold X = Add/Remove Favourite**.
- If the new Hold Y binding is absent and the legacy `hold_y_random` value is true, resolve it as **Hold Y = Random Game**.
- False or absent legacy values resolve to **None**.
- An explicitly saved new binding takes precedence over a legacy value.
- On the next successful settings save, write the resolved new bindings. Keep the legacy X/Y values consistent with their equivalent mappings so returning to an older Degauss release does not accidentally enable the wrong shortcut or lose the two behaviours that release understands.
- Existing settings files without shortcut values continue to behave as before, with all four shortcuts disabled.

### Scope

Do not add Start, Select, configurable hold duration, short-press reassignment, arbitrary controller remapping or additional actions. Scraping, rebuilding, hiding items and changing core, image or game-data settings remain in Actions because they are consequential or multi-step operations.

### Acceptance criteria

- All four face buttons can be independently assigned every supported shortcut action.
- A normal tap and a one-second hold are distinguished correctly for A, B, X and Y.
- A completed hold never also launches, opens, goes back, opens Actions or opens General Options on release.
- Context-dependent shortcuts act only where their existing Actions equivalent is available.
- An unavailable shortcut leaves the button's ordinary action working normally.
- Cycle View is saved only for the exact current place and survives leaving it, restarting Degauss and rebooting MiSTer.
- Random shortcuts preserve both existing Random Game Behaviour modes.
- Existing Hold X and Hold Y settings load with the same enabled behaviour after upgrading.
- Saving and reopening Options preserves every assignment.
- Controller testing on MiSTer covers taps, holds, cancellation by movement, game launch, back navigation, every browsing level and contexts where each shortcut is unavailable.

giancarloerra:
## What this changes

Adds a dedicated Shortcuts options page with independent one-second hold assignments for A, B, X and Y. Each button can run Cycle View, Random Game, Random Favourite, Add/Remove Favourite, Game Information, Search This Folder or Jump to Letter. Short presses retain their existing actions, and unavailable shortcuts leave the normal press immediate.

Existing Hold X favourite and Hold Y random settings migrate without resets and remain compatible when settings are next saved.

## Why

Closes #78.

The previous fixed Hold X and Hold Y options could not assign the other common browsing actions or use A and B.

## How it was checked

- [x] `cargo test`: 713 passed, 5 credential-dependent live tests intentionally ignored
- [x] `cargo fmt --check`
- [ ] Run on a real MiSTer

Additional checks: all 10 upgrade rehearsals passed; the Shortcuts page was rendered and inspected at 352x240 with both default and longest labels.

## Licensing

- [x] I agree


giancarloerra:
Superseded by and included in #82, which has now merged into main.

PR #89, opened 2026-09-16
Add temporary metadata game filters

### LINKED ISSUE #83: Add temporary metadata filters to game folders
## Problem

`Search This Folder` can narrow a game list by title, but metadata already loaded for those games cannot be used to narrow the list. Large libraries would be easier to explore with temporary filters such as genre, year or player count.

## Required interface

Add **Filter Games** under **Actions -> Find** while browsing games inside a system or folder.

The filter screen contains these rows:

- Genre
- Year
- Players
- Language
- Developer
- Publisher

Each row initially shows **Any**. Pressing **A** opens a normal list containing:

1. **Any**
2. every distinct non-empty value present in the unfiltered playable rows of the current folder;
3. **Unknown**, only when at least one playable row has no value for that field.

If the current source supplies no values for a field, show **Unavailable** and do not open an empty value list. Choosing a value applies it and returns to the filter screen. **B** returns to browsing and keeps the selected filters.

When at least one metadata filter is active, add **Clear Filters** under **Actions -> Find**. It clears metadata filters only.

## Filtering behaviour

- One value can be selected per field.
- Active fields combine with AND.
- Metadata filters and `Search This Folder` combine with AND.
- **Clear Search** retains metadata filters; **Clear Filters** retains the title search.
- Compare trimmed values case-insensitively while showing their original text.
- Derive Year from the four-digit year already presented by Degauss. Other fields use their complete stored value without inventing semantic categories.
- Apply a value only after it is chosen with **A**. Moving through a value list must not repeatedly rebuild the visible game list.
- Filter playable games only. Keep subfolders visible so nested libraries remain navigable.
- Filters belong only to the current browse place. Entering another folder, leaving the system, rebuilding the list or restarting Degauss clears them, matching the existing place-specific search behaviour.
- If Random Game or Random Favourite is used while filters are active, choose only among the visible matching games in the current folder. Their existing recursive behaviour remains unchanged when no filter is active.
- Launching, Game Information and favourite actions on a matching row retain their existing behaviour.

Filter values must always be derived from the complete unfiltered rows for the current folder, not from the result of another active filter. This lets a user change one criterion without values disappearing because of another criterion.

## Metadata sources and special collections

- Filter the effective metadata currently shown by Degauss, whether it came from `gamelist.xml` or a selected MiSTer Artwork Pack.
- A source that does not provide a particular field simply makes that field unknown or unavailable. Do not invent values or consult another source.
- Favourites use the metadata resolved from their original system and current selected data source.
- Games inside ZIP libraries use the same filtering behaviour as ordinary files.
- Description is excluded because it is free text and complete descriptions are loaded on demand. Names remain covered by `Search This Folder`.
- Rating is excluded because it is not currently part of Degauss's metadata model.

## Performance and compatibility

- Filter only rows already held in memory.
- Do not read directories, parse metadata files, contact a provider, rebuild an index or change a cache while applying a filter.
- Do not change the cache format, settings format, system table or existing metadata-source behaviour.
- Do not persist filters in Settings or require migration, reset or reindexing.
- With no active filters, browsing and Random Game behaviour remain unchanged.
- Existing libraries with partial or no metadata remain usable.

## Acceptance checks

- A folder containing synthetic games with different genre, year, players, language, developer and publisher values offers the correct distinct choices.
- Each field filters independently, and two or more fields combine with AND.
- `Search This Folder` combines with metadata filters; clearing either one leaves the other active.
- Games with missing metadata appear only when the corresponding field is **Any** or **Unknown**.
- Subfolders remain visible while their sibling games are filtered.
- Entering another folder and returning clears the previous place's filters.
- Random Game and Random Favourite cannot select a game excluded by an active filter.
- Gamelist, Artwork Pack, Favourites and ZIP-member rows use their effective displayed metadata.
- Existing launch, information, favourite, search, jump, hide and view actions still operate on the correct selected row.
- At 352x240 and 640x480, every field and value remains readable and selectable.
- On MiSTer, applying and clearing filters on a full Arcade list remains responsive and performs no indexing or storage activity.

giancarloerra:
Closes #83

## Summary
- add Actions → Find → Filter Games for genre, year, players, language, developer and publisher
- combine temporary metadata criteria with title search while keeping folders reachable
- scope filtered random actions to visible games and keep filters out of settings and cache files
- document the controls and preserve existing title-search relist behaviour

## Validation
- cargo fmt --all --check
- cargo clippy --target armv7-unknown-linux-musleabihf --all-targets -j 1 -- -D warnings
- cargo test
- CodeRabbit CLI review: 0 findings
- software-renderer acceptance at 352×240 and 640×480

## Summary by CodeRabbit

- **New Features**
  - Added game filtering by genre, year, number of players, language, developer and publisher.
  - Filters can be combined with title searches and applied to random game selection.
  - Added options to clear filters and identify unavailable metadata values.
  - Filters remain active when refreshing the current folder and clear when changing browsing context.

- **Bug Fixes**
  - Improved preservation of browsing context and filters during relisting.


giancarloerra:
Closed by #103.

PR #91, opened 2026-09-16
Add an optional deduplicated Last Played collection

### LINKED ISSUE #84: Add an optional deduplicated Last Played collection
## Purpose

Add an optional **Last Played** collection containing the most recently launched games, without requiring a resident process or changing Degauss's exit-on-launch architecture.

## Option and top-level entry

Add **Last Played** to **Options -> Library** with these values:

- Off
- 1 through 10

An absent setting means **Off**, preserving existing installations. When set to a number:

- show **Last Played** as a top-level entry immediately before Favourites;
- opening it goes directly to its mixed-system game list, without an intermediate system row;
- show at most the selected number of games, newest first;
- if nothing has been launched yet, show a clear empty-state message;
- **B** returns directly to the top-level categories.

Changing the displayed amount does not discard newer retained history. Degauss retains at most ten unique entries. **Off** hides the entry and stops recording new launches without deleting already retained history.

## Recording and deduplication

Record a game only on the committed launch path, after validation and launch-plan construction have succeeded. A missing game, missing core, invalid ZIP member or other launch error that leaves Degauss running must not enter the history.

Degauss cannot determine how long a core was used because it exits when handing the launch to MiSTer. For this feature, “played” therefore means that Degauss successfully handed off the launch.

Identify an entry by its originating system and exact logical launch target:

- ordinary files use their exact target path;
- ZIP games include the complete archive-member identity;
- AmigaVision entries include the installation and title;
- a game launched from Favourites resolves to its original game and system before recording;
- launching the same logical game from its system, Favourites or Last Played produces one entry;
- the same file intentionally launched through different systems remains distinct because its launch configuration can differ.

Launching an existing entry moves it to the front rather than adding a duplicate. Trim the oldest entry only when the retained list exceeds ten.

## Display and behaviour

Store only the stable launch identity, originating system and last known display name. Do not persist copied artwork or metadata.

Whenever Last Played is opened:

- resolve each entry against its originating system's current cached rows and selected metadata source;
- show the current game name, artwork, metadata and full Game Information exactly as the originating system would;
- show current favourite state and retain Add/Remove Favourite behaviour;
- launch through the originating system's current core selection and launch rules;
- support ordinary files, MRA, MGL, ZIP members, AmigaVision and other launchable rows already supported by Degauss;
- retain the recorded display name when current presentation metadata is unavailable, while leaving unavailable artwork and metadata empty;
- use the most recent resolvable game's current artwork for the top-level preview unless an existing custom category image for Last Played is selected.

Treat Last Played as one stable browse place. It follows the global view by default and can receive its own place-specific view through the existing Change View action. Search, Jump and other game-list actions that are meaningful for a mixed-system collection must continue to target the correct row and originating system. System-wide actions such as rebuilding a system or choosing one system's source must not pretend the mixed collection has a single system.

If a recorded target has since disappeared, keep the history entry and use the existing launch-time missing-file error. Do not silently delete history or scan the filesystem while opening the collection.

## Storage, performance and errors

- Keep launch history in a dedicated, versioned file beside Degauss's other runtime data, separate from user Settings and the temporary return-position state.
- Use a bounded, atomic write so a power interruption cannot leave a partially written history file.
- A missing history file is the normal empty state.
- A malformed history file or write failure must be reported clearly without preventing ordinary browsing or a requested game launch.
- Opening Last Played may load only the system caches needed by its maximum ten entries, grouping entries by system so each cache or prepared Artwork Pack provider is reused.
- Do not scan directories, rebuild an index, copy artwork, create favourite files or add a background process.

## Backward compatibility

- Existing installations have the feature disabled until explicitly enabled.
- Existing Settings, return state, caches, Favourites and launch paths remain compatible and require no migration or rebuild.
- With the option Off, the top-level menu and launch behaviour remain unchanged.
- Removing, renaming or changing metadata for a game does not corrupt the history file or another system's cache.
- Existing selected standard, RetroAchievements or Unstable core behaviour is resolved at launch time rather than copied into history.

## Acceptance checks

- With the option Off, no Last Played entry appears and successful launches do not change history.
- Values 1 through 10 show exactly that maximum number of retained games.
- Eleven different successful launches retain only the newest ten; lowering the visible amount and raising it again preserves the retained order.
- Replaying an existing game moves it to the front without duplication.
- Launching the same game from its original system and from Favourites still produces one entry.
- Failed preflight validation does not create or reorder an entry.
- Ordinary files, MRA, MGL, ZIP members and AmigaVision entries relaunch correctly.
- Gamelist and Artwork Pack presentation changes appear in Last Played without rewriting history.
- Favourite markers and Add/Remove Favourite update immediately and agree with the original system.
- A missing target stays visible and produces the existing actionable missing-file error when selected.
- Last Played follows the global view, saves an explicit custom view for only this collection and returns directly to the top level with **B**.
- Restarting Degauss and rebooting MiSTer preserve order and deduplication.
- Opening the collection performs no index rebuild or directory scan and remains immediate on MiSTer.

giancarloerra:
## Summary

- add an optional Last Played collection with a configurable 1 to 10 game limit
- record only successfully committed launches and deduplicate by originating system and exact launch identity
- resolve current artwork, metadata, favourite state and launch rules when the collection is opened
- persist bounded history atomically while preserving existing settings, caches and launch behaviour when disabled

Closes #84

## Summary by CodeRabbit

* **New Features**
  * Added an optional **Last Played** collection showing up to 10 recently launched games, newest first.
  * Added Library settings to configure the number of visible entries or disable the collection.
  * Last Played entries support filtering, favourites, artwork, game information, random selection and relaunching.
  * History is restored after restarting and updated following successful launches.
  * Launches retain their originating system for reliable relaunching.

* **Bug Fixes**
  * Improved handling of missing or invalid history data with clear startup and empty-view messages.
  * Preserved launch behaviour when history cannot be saved.


giancarloerra:
Closed by #103.

PR #99, opened 2026-09-16
Support AmigaVision April 2026 direct launches

### LINKED ISSUE #97: Support direct launch from AmigaVision April 2026 friendly listings
## Summary

Support direct game launching with `AmigaVision-MiSTer-2026.04.26-r2.7z` while preserving the friendly game names shown in Degauss.

That AmigaVision package changed `listings/games.txt` to contain friendly display names. AmigaVision's external-launch interface still resolves only exact canonical names from `AGS:RunQuiet/Game`. Degauss currently uses the selected `games.txt` line as both the displayed title and the value written to `ags_boot`, so a selected game can open the AmigaVision interface instead of launching directly.

This should be handled transparently by Degauss. Users should not need to replace AmigaVision files, edit listings, rename games, or rebuild their Degauss library.

## Current behaviour

The current launch path is:

1. Read game names from AmigaVision's `listings/games.txt`.
2. Show those names in the Amiga browser.
3. Write the selected name to `<AmigaVision install>/shared/ags_boot`.
4. Start the Amiga core.

With the affected package, step 3 writes a friendly display name. AmigaVision's `ags-launch` checks for an exact file with that name under `AGS:RunQuiet/Game` or `AGS:RunQuiet/Demo`. If no exact entry exists, no game is executed and AmigaVision continues into its normal interface.

The public upstream report provides this example:

```text
listings/games.txt:       Superfrog
AGS:RunQuiet/Game entry:  Superfrog (CD32)[en]
```

Writing `Superfrog` to `ags_boot` therefore cannot select `Superfrog (CD32)[en]`.

## Verified upstream cause

AmigaVision generates the two sides of the interface differently:

- [`generate_listings.py`](https://github.com/amigavision/AmigaVision/blob/bb55143350c539b81caa1d9a8364639685ea4d33/build/generate_listings.py#L19-L37) builds `games.txt` from the friendly browse-tree filenames.
- [`ags_make.py`](https://github.com/amigavision/AmigaVision/blob/bb55143350c539b81caa1d9a8364639685ea4d33/build/ags_make.py#L67-L84) builds canonical names by adding hardware and language information.
- [`ags-launch`](https://github.com/amigavision/AmigaVision/blob/bb55143350c539b81caa1d9a8364639685ea4d33/configs/Layers/Base/DH0/S/ags-launch#L18-L24) performs exact lookup using the value received through `ags_boot`.

The same incompatibility is documented in [AmigaVision issue #465](https://github.com/amigavision/AmigaVision/issues/465) and [Zaparoo issue #1374](https://github.com/ZaparooProject/zaparoo-core/issues/1374).

## Required behaviour

For the exact affected AmigaVision listing version:

1. Keep showing the friendly title from `games.txt` in Degauss.
2. Resolve that friendly title to its exact canonical `AGS:RunQuiet/Game` target only when launching.
3. Write the canonical target to `ags_boot`.
4. Start AmigaVision through the existing launch path.

For example:

```text
Displayed in Degauss:  Superfrog
Written to ags_boot:   Superfrog (CD32)[en]
```

All other AmigaVision versions and custom listings must retain their existing behaviour unless an exact, separately supported mapping is available.

### 1. Add an exact, version-bound compatibility map

Generate a display-name to canonical-target map from the public affected AmigaVision package. The source of each mapping must be the package's actual run entries, not a heuristic that removes or invents suffixes.

The compatibility data should be:

- tied to the exact affected `games.txt` content;
- deterministic and checked into the repository;
- line-safe for the listing's Latin-1 encoding;
- limited to direct game targets that AmigaVision can resolve;
- validated so duplicate display names, duplicate targets, missing rows, or malformed rows fail during tests rather than producing an arbitrary launch.

The runtime must not mount or parse the PFS3 HDF. The mapping is derived once from the public package and used as static compatibility data.

### 2. Recognise only the exact affected listing

Before applying the map, identify `listings/games.txt` by an exact stable fingerprint of its bytes.

Use a cheap file-size check before reading and hashing the candidate file. Apply the compatibility map only when the complete fingerprint matches.

This prevents an older release, a future release, or a user-edited listing from being interpreted using the wrong mapping. Similar-looking filenames or titles are not sufficient evidence.

### 3. Resolve at the common AmigaVision launch boundary

Perform the translation in the shared AmigaVision launch planner immediately before creating the `ags_boot` launch plan.

This ensures the same correction applies to:

- a game selected from the Amiga browser;
- an existing Degauss favourite that refers to that AmigaVision game;
- any other existing Degauss route that reaches the same AmigaVision launch function.

Do not copy the mapping logic into individual browse or favourites paths.

### 4. Preserve display and stored identity

The translated canonical name is a launch-only value.

Do not change:

- the title shown in Degauss;
- cached game names or cache format;
- favourites or their stored identity;
- artwork or metadata lookup names;
- search, sorting, filtering, or navigation;
- the user's `games.txt` or any AmigaVision content;
- the indexing format or indexing workflow.

No cache rebuild, migration, reset, or user action should be required.

### 5. Keep demos unchanged

This issue concerns the game listing generated from the friendly all-games browse tree. `demos.txt` is generated from AmigaVision's run directory by the upstream script and must remain on the existing path unless a separate exact incompatibility is demonstrated.

### 6. Handle AmigaVision known-issue entries explicitly

Some friendly game-list entries lead to AmigaVision's own known-issue launchers rather than a direct `RunQuiet/Game` target. They must not be guessed, silently redirected, or allowed to fall through to the AmigaVision interface as if launch succeeded.

When such a title is selected, show an actionable message such as:

```text
This title is listed by AmigaVision under known issues and cannot be launched directly. Open it from the AmigaVision menu instead.
```

If a recognised affected listing contains a title absent from the checked compatibility data, fail with a distinct mapping error. Do not write the friendly title and present the resulting AmigaVision interface as a successful direct launch.

## Why not replace `games.txt`

Shipping a replacement listing would modify another project's installation and make the visible titles depend on which copy happened to be installed. It could also change Degauss game identity and require a reindex.

A launch-only resolver has the smaller and more compatible behaviour surface:

- AmigaVision remains untouched.
- Degauss retains the friendly names users installed.
- existing Degauss data remains valid;
- only the value sent across the broken launch boundary changes.

## Why not derive canonical names heuristically

Hardware and language suffixes are not safely recoverable from a friendly title alone. A title may have multiple editions or targets, and punctuation or non-ASCII characters may be significant. Stripping or appending text would create ambiguous and potentially incorrect launches.

Only an exact mapping derived from the matching public package is acceptable.

## Backward compatibility requirements

- Existing AmigaVision installations whose listings already contain canonical names must behave exactly as before.
- Older, newer, custom, modified, missing, or unreadable `games.txt` files must not be force-mapped as the affected package.
- A one-byte change to the affected listing must prevent the version-specific mapping from being selected.
- Existing Degauss favourites must continue to resolve without migration.
- Existing game titles, artwork, metadata, ordering, filtering, and cache identities must remain unchanged.
- Existing demo launching must remain unchanged.
- The current `ags_boot` file location and core-launch sequence must remain unchanged.
- No new persistent process, service, background scan, or installation step may be introduced.

## Performance requirements

The compatibility check must not add work to normal indexing or browsing.

- Do not open or hash AmigaVision data while scanning unrelated systems.
- Do not inspect the HDF at runtime.
- Check the listing only when an AmigaVision game is launched.
- Reject clearly different files using metadata before reading their contents.
- Do not add cache transactions, reindexing, filesystem traversal, or validation to ordinary library paths.

## Acceptance criteria

- [ ] The affected public AmigaVision package still displays its friendly game names in Degauss.
- [ ] Selecting a normal affected game writes the exact canonical target expected by `AGS:RunQuiet/Game`.
- [ ] The selected game launches directly instead of opening the AmigaVision interface.
- [ ] A title with multiple possible editions resolves to the exact package-defined target, not a guessed target.
- [ ] A title containing Latin-1 characters resolves byte-for-byte correctly.
- [ ] An existing favourite for an affected game uses the same correction and launches directly.
- [ ] A recognised AmigaVision known-issue entry produces a clear, specific message and does not report a successful direct launch.
- [ ] A recognised listing row missing from the bundled map produces a clear mapping-integrity error.
- [ ] A demo continues through the unchanged demo launch path.
- [ ] An older canonical `games.txt` entry is passed through unchanged.
- [ ] A custom or one-byte-modified `games.txt` is not mapped using the affected release's table.
- [ ] Missing or unreadable listing data preserves the existing actionable error behaviour.
- [ ] No cache or configuration migration is required.
- [ ] No library reindex is required after upgrading Degauss.
- [ ] Indexing and browsing perform no new AmigaVision compatibility work.

## Automated tests

Add tests covering:

1. Exact fingerprint recognition for the affected public listing.
2. Rejection of a same-sized listing with different content.
3. Rejection of shorter, longer, older, future, and custom listings.
4. Exact friendly-name to canonical-target resolution.
5. A title with ambiguous editions.
6. Latin-1 title bytes and punctuation.
7. Existing canonical title pass-through.
8. Demo pass-through.
9. Existing-favourite launch through the common resolver.
10. Explicit known-issue classification.
11. Explicit failure for a recognised listing whose map is incomplete.
12. Static validation that every compatibility-map row is well formed and unique.
13. Static validation that the map corresponds exactly to its recognised listing version.
14. Preservation of existing cache, favourite, metadata, artwork, and launch-plan serialization behaviour.

## On-device validation

Using the public affected AmigaVision package on a MiSTer:

1. Launch a representative normal game from the Degauss Amiga browser and confirm that the selected game boots directly.
2. Launch the same game from an existing Degauss favourite and confirm the same result.
3. Select an AmigaVision known-issue entry and confirm that Degauss gives the specific explanation without entering an apparent successful launch path.
4. Restore an older or custom listing and confirm its previous launch behaviour is unchanged.
5. Confirm that entering, searching, browsing, and indexing the Amiga library show no new delay or rebuild requirement.

## Non-goals

- Changing AmigaVision's package or upstream generator.
- Downloading or replacing the user's listing files.
- Parsing or mounting the AmigaVision HDF at runtime.
- General fuzzy matching of AmigaVision titles.
- Guessing hardware, language, edition, or region suffixes.
- Making AmigaVision's own known-issue entries directly launchable.
- Changing demo handling without separate evidence.
- Redesigning the general Degauss launch-rule system.

giancarloerra:
## Summary

- recognize the exact affected AmigaVision April 2026 games listing
- preserve its friendly display names while translating them to canonical launch names only at launch time
- keep older, future, customized, demo, cached, and favourite paths unchanged
- report the package's known non-launchable entries clearly instead of falling through to the AmigaVision UI

## Compatibility and performance

- no user files are rewritten
- no reindex or cache migration is required
- unaffected installations perform only a cheap listing-size check before keeping the existing launch path
- the full listing hash and static mapping are used only for the exact affected package

## Validation

- ARM release build passed
- ARM Clippy passed with warnings denied
- 918 tests passed; 5 existing tests remained intentionally ignored
- mapping completeness, uniqueness, accented titles, ambiguous editions, demos, modified listings, favourites, and known unsupported entries are covered
- affected and older listing paths were exercised on MiSTer

Fixes #97

## Summary by CodeRabbit

- **New Features**
  - Added support for the April 2026 AmigaVision launch catalogue.
  - Boot titles from the recognised games listing now use canonical launch names where available.
  - Unrecognised or unchanged titles continue to be preserved.

- **Bug Fixes**
  - Known-issue entries are now rejected.
  - Invalid or incomplete catalogue mappings now produce clear errors instead of incorrect launch titles.
  - Existing Latin-1 encoding and newline handling remain supported.


giancarloerra:
Closed by #103.

PR #102, opened 2026-09-17
Show one-game ZIP archives directly as games

giancarloerra:
## What this changes

- Shows the sole supported member of a generic ZIP directly in its containing folder.
- Preserves exact archive-member launch targets, metadata, artwork, Favourites and saved-position recovery.
- Keeps multi-game ZIPs, whole-archive systems, Neo Geo sets and malformed-archive handling unchanged.
- Reuses the archive read performed by cache construction and removes unreachable virtual ZIP screens from the published cache.

## Why

One-game ZIP libraries currently require opening a redundant archive folder before launching the only game. Closes #100.

## How it was checked

- [x] `cargo test`
- [x] `cargo fmt --check`
- [x] `cargo clippy --target armv7-unknown-linux-musleabihf --all-targets -- -D warnings`
- [x] Run on a real MiSTer
- [ ] On a CRT, if it changes anything drawn

## Licensing

By opening this pull request I agree that my contribution is licensed under
the [PolyForm Noncommercial License 1.0.0](../LICENSE), the same licence as
the rest of Degauss, and that I have the right to license it that way.

Changes to `MiSTer_Degauss` do not belong here: it is a separate GPLv3
program in its own repository,
[Degauss-Main](https://github.com/giancarloerra/Degauss-Main).

- [x] I agree

## Summary by CodeRabbit

- **New Features**
  - Single-game ZIP archives are displayed directly as playable games in the containing folder.
  - Direct game presentation includes artwork and metadata across Details, Tiled, Carousel and Gallery views.
  - Nested single-game archives and supported companion files are handled consistently.
  - Multi-game archives remain accessible as virtual folders, while systems that launch ZIP files retain whole-archive behaviour.

- **Bug Fixes**
  - Improved navigation and saved-position restoration when archive contents are displayed directly.
  - Archive presentation changes are reflected during library rebuilds and artwork or scraper operations.

PR #96, opened 2026-09-16
Let each system choose among compatible installed cores

giancarloerra:
## Summary

- add a per-system **Launch Core** selector for declared compatible core families
- keep Automatic as the default, with explicit profile IDs persisted separately from Core Version
- apply the selected family to normal games, ZIP members, new favourites and existing favourites
- keep unavailable explicit selections visible and fail clearly instead of silently falling back
- preserve existing systems tables and settings without migration or index rebuilds

## Validation

- `cargo fmt --check`
- ARM `cargo clippy --target armv7-unknown-linux-musleabihf --all-targets -j 1 -- -D warnings`
- `cargo test` (923 passed, 5 ignored)
- `./scripts/rehearse-migration.sh` (10/10 passed)
- CodeRabbit CLI review completed; its sole finding duplicated validation already enforced and tested at systems-table load
- Snyk Code reported only the existing low-severity findings in unchanged maintenance scripts before the account returned 403
- Snyk dependency scan reported no supported target files

Stacked on #95 so this PR contains only the Launch Core change. Retarget to `main` after #95 merges.

Closes #93

## Summary by CodeRabbit

* **New Features**
  * Added a context-menu option to choose from compatible installed core families before launching a game.
  * Supports automatic selection, explicit choices, saved preferences, and core-version selection.
  * Launches and favourites now honour the selected core family.
  * Displays unavailable selections and reports relevant save or launch errors.

* **Bug Fixes**
  * Improved core discovery for Standard, RetroAchievements, and Unstable versions.
  * Preserved existing behaviour for self-describing game files and older settings.

PR #95, opened 2026-09-16
Support current Neo Geo Pocket cores

giancarloerra:
## Summary

- use the current `_Console/NGPC` core for Neo Geo Pocket and Neo Geo Pocket Color
- retain automatic compatibility with the legacy JTNGP and JTNGPC cores
- support shared NGPC libraries and the `.npc` extension without changing existing system identities
- preserve launch, favourites, RA and Unstable-core behaviour across compatible core families

## Validation

- formatting and ARM Clippy
- full test suite
- installation and migration rehearsal
- shipped and deployed systems-table identity
- CodeRabbit local review

Closes #92

## Summary by CodeRabbit

* **New Features**
  * Added support for compatible core families, enabling preferred and legacy cores to be used when available.
  * Improved core selection, launch resolution and library matching across compatible core configurations.
  * Neo Geo Pocket and Neo Geo Pocket Color now support additional folders, file extensions and legacy core compatibility.
  * Neo Geo Pocket Color launches now support both `.ngc` and `.npc` game files.
* **Bug Fixes**
  * Improved recognition of legacy Neo Geo Pocket entries and associated game libraries.
  * Added validation for compatible core definitions.

PR #103, opened 2026-09-17
Prepare Degauss 0.8.0

### LINKED ISSUE #83: Add temporary metadata filters to game folders

### LINKED ISSUE #84: Add an optional deduplicated Last Played collection

### LINKED ISSUE #85: Add an optional cached Cores browser
## Summary

Add an optional **Cores** entry to the main Degauss screen, beside Arcade, Console, Computer and Favourites. It provides a cached browser for launching installed MiSTer cores without selecting a game.

This is the Degauss equivalent of a direct core launcher, while preserving Degauss's existing exit-on-launch and return behaviour.

## User interface

Add **Options → Library → Show Cores**, default **Off**.

When enabled, the main screen includes **Cores**. Opening it first shows the relevant categories found on the card:

```text
Cores
├── Console
├── Computer
├── Utility
├── Other
├── Custom menu folders
└── Unmatched Unstable
```

Each category contains one row for every installed launchable variant, sorted alphabetically by logical system name and then by variant in this order: Standard, RA, Unstable.

For example:

```text
Console
├── Game Boy [Standard]
├── Game Boy [RA]
├── Mega Drive [Standard]
├── NES [Standard]
├── NES [RA]
├── NES [Unstable: 20260912]
├── SNES [Standard]
└── SNES [RA]
```

The variant suffix must always be visible in text. It must not depend on colour or theme. A Details-style view may use an existing system/core logo and show the full variant or build name when the list truncates it.

Pressing **A** launches the selected core without loading a game. **B** returns to the previous level.

## Core classification and launch targets

- **Standard** entries are visible installed `.rbf` cores.
- **RA** entries come from validated `_RA_Cores/*.mgl` launchers and launch through those files so the required RA configuration is retained.
- The supporting `.rbf` files under `_RA_Cores/Cores` are internal and must not appear as duplicate rows.
- Matching Unstable builds inherit the Console, Computer or custom category of their logical system and retain an identifiable build name.
- An Unstable core that cannot be matched safely appears under **Unmatched Unstable**.
- `_Arcade/cores` contains support files and must not be exposed. Arcade games remain in the existing Arcade browser and launch through their `.mra` definitions.
- Other arbitrary `.mgl` and `.mra` files are not part of this core browser.

## Cached discovery and rebuilding

Opening **Cores** must never enumerate the card.

Degauss already discovers installed cores while preparing its system list. Extend that existing discovery result to produce a persisted Cores catalogue in the same pass, without a second traversal.

- **Index All Systems** refreshes the Cores catalogue as part of its existing system/core discovery stage.
- Opening and re-entering Cores reads the completed catalogue only.
- On an installation without a Cores catalogue, the result of the discovery Degauss already performs supplies it. A full game-library rebuild must not be required solely to introduce this feature.
- Inside Cores, **Actions → Library → Rebuild Cores List** replaces **Rebuild This System List** and scans only the shallow MiSTer menu/core folders.
- Rebuilding Cores must not scan game folders, ZIP contents, artwork, gamelists or metadata.
- A successful targeted rebuild immediately refreshes the displayed Cores catalogue.

The catalogue must be additive to existing persisted data. Existing game caches and settings must remain valid.

## Available Actions

The Cores browser is not a game library. It may offer only operations that apply to cores:

- Launch
- Search by core name
- Jump to letter
- Change View / Use Global View
- Rebuild Cores List

Do not offer metadata filters, Game Information, scraping, game-data sources, artwork-pack selection, random-game actions, Favourites or Last Played actions.

## Backward compatibility

- The entry is hidden by default, so existing main screens remain unchanged.
- Existing Standard/RA preference and per-system Core Version selections remain unchanged. The Cores browser launches the explicitly selected row and does not modify those settings.
- Existing game, favourite, Arcade, Other, Utility and Unstable browsing behaviour remains unchanged.
- Custom `menu_root` and custom menu folders remain supported.
- Older settings and cache files remain readable without a migration or reset.

## Acceptance criteria

- Enabling Show Cores adds Cores to the main screen; disabling it removes only that entry.
- Repeatedly opening Cores performs no directory enumeration.
- Standard, RA and matching Unstable variants appear together under their logical category with unambiguous text labels.
- RA-only systems appear as `[RA]` and launch through their validated RA launcher.
- Internal RA and Arcade support cores never appear as duplicates.
- Unmatched Unstable cores remain available without being guessed into the wrong category.
- A selected Standard, RA and Unstable example each launches successfully on MiSTer and Degauss returns normally after exiting the core.
- Rebuild Cores List detects an added and a removed core without rebuilding any game system.
- Index All Systems refreshes the same catalogue without performing an additional core traversal.
- Existing main-menu categories, game launching and saved core preferences behave exactly as before when Show Cores is Off and On.
- Compare startup and Index All performance on the same MiSTer before and after the change. There must be no material regression, and opening the cached Cores browser must be immediate.

### LINKED ISSUE #86: Add display-only game name formatting and optional folder brackets
## Problem

Filename-derived game names can spend scarce horizontal space on region, version, translation and dump tags. For example:

```text
Seiken Densetsu 3 (Japan) [T+ENG]
```

Degauss also adds an outer `[ name ]` marker to folder rows. This is separate from brackets contained in a title, but libraries where each game has its own folder can therefore appear to wrap every game name.

### Game Name Display

Add a global `Game Name Display` option under Appearance with the following modes:

- `Full` (default)
- `Remove Parentheses`
- `Remove Brackets`
- `Remove Parentheses and Brackets`
- `Keep Region`
- `Keep Disc Index`
- `Keep Region and Disc Index`

These follow RetroArch's playlist label-display choices and retain useful ways to distinguish regional and multi-disc entries: https://github.com/libretro/RetroArch/blob/d7d75f4000bb748f664465ef949ef261e3ed1698/playlist.h#L49-L57

For the example above:

| Mode | Visible name |
| --- | --- |
| Full | `Seiken Densetsu 3 (Japan) [T+ENG]` |
| Remove Brackets | `Seiken Densetsu 3 (Japan)` |
| Remove Parentheses | `Seiken Densetsu 3 [T+ENG]` |
| Remove Parentheses and Brackets | `Seiken Densetsu 3` |

Apply the selected mode to the final effective name after filename, `gamelist.xml`, or Artwork Pack name resolution. Apply it consistently to list, details, tile and Favourites presentation. Visible sorting and jump-to-letter must use the displayed name rather than a hidden prefix or tag.

This must be a runtime presentation transform only:

- Do not alter filenames, folders, launch paths or favourite targets.
- Do not edit `gamelist.xml` or Artwork Pack data.
- Keep the original full effective name in the cache/index.
- Do not persist a shortened replacement name anywhere.
- Changing the option must update the current library from its existing data without rescanning, reindexing or rewriting a cache.

### Folder Brackets

Add a separate global `Folder Brackets` option under Appearance:

- `On` (default)
- `Off`

This controls only Degauss's outer `[ name ]` decoration for folder rows. It must not remove square-bracketed content from the underlying title. Apply game-name formatting before optionally adding the folder marker.

## Compatibility

- Existing configurations without these settings retain the current full names and folder brackets.
- Launching, favourites, hidden entries, scraping and metadata matching continue to use the existing path and source identity.
- Visually identical names produced by filtering remain distinct entries backed by their original paths.
- No cache-format migration or library rebuild is required.

## Acceptance checks

- Verify every display mode with filename-derived, `gamelist.xml` and Artwork Pack names.
- Confirm the stored XML, Artwork Pack source, cache/index name and filesystem entry remain unchanged after changing modes and restarting Degauss.
- Confirm list, details, tile and Favourites views show the same transformed name.
- Confirm sorting and jump-to-letter follow the transformed visible name.
- Confirm `Folder Brackets` affects only Degauss's outer folder marker.
- Confirm games with identical shortened labels still launch and favourite the correct original targets.
- Confirm changing either option causes no scan, reindex or cache rewrite.
- Confirm an existing configuration with neither setting renders exactly as before.

### LINKED ISSUE #92: Use the current NGPC core for Neo Geo Pocket libraries
## Confirmed problem

Degauss v0.7.1 still maps its two Neo Geo Pocket systems to Jotego's older cores:

```toml
name = "Neo Geo Pocket"
rbf = "_Arcade/JTNGP"
setname = "NeoGeoPocket"

name = "Neo Geo Pocket Color"
rbf = "_Arcade/JTNGPC"
setname = "JTNGPC"
```

Source: https://github.com/giancarloerra/Degauss/blob/b94df54e17c2061c206f83b7e4782ce36b7101a8/assets/systems.toml#L1106-L1133

MiSTer-devel now provides `NGPC_*.rbf` as the current core. It lives under `_Console`, supports both monochrome Neo Geo Pocket and Neo Geo Pocket Color, uses `games/NGPC`, and accepts `NGP`, `NGC` and `NPC` cartridge extensions:

- https://github.com/MiSTer-devel/NGPC_MiSTer/blob/238299fd4c128406d506daa787e5b6abf92915ad/README.md#L1-L7
- https://github.com/MiSTer-devel/NGPC_MiSTer/blob/238299fd4c128406d506daa787e5b6abf92915ad/NGPC.sv#L84-L88

When the JT cores have been removed and `_Console/NGPC_*.rbf` is installed, Degauss does not recognise `NGPC` as the replacement for either configured JT core. It therefore leaves the systems under Arcade and generated launches still name a missing JT core.

## Required behaviour

- Treat `_Console/NGPC` as the current standard core for both Neo Geo Pocket and Neo Geo Pocket Color games.
- Place those systems under Console, or under the optional Handheld category when that category is enabled.
- Launch `.ngp`, `.ngc` and `.npc` cartridges through `_Console/NGPC` using file index `1`.
- Do not emit the JT-specific set names when launching through the current core; its cartridge extension selects the required mode.
- Preserve the existing `NeoGeoPocket` and `NeoGeoPocketColor` logical IDs so caches, metadata, artwork packs, scraper mappings, favourites and saved per-system settings retain their identities.

## Existing library compatibility

- Existing `games/NGP` libraries containing `.ngp` files must continue to work.
- Existing `games/NGPC` libraries containing `.ngc` files must continue to work.
- Follow the current core's documented single-folder arrangement by also accepting `.ngp`, `.ngc` and `.npc` files from `games/NGPC`.
- If the current `NGPC` core is absent but the corresponding JT core is still installed, preserve the existing JT launch as a legacy fallback. If both are installed, prefer the current MiSTer-devel core.
- Existing indexed `.ngp` and `.ngc` games must not require a full-library rebuild or cache migration. A targeted rebuild may be needed only to discover an extension that the older table ignored, such as `.npc` or `.ngp` newly placed in `games/NGPC`.
- Custom `systems.toml` files must remain valid and must not be silently overwritten.

## Delivery

- Correct the shipped systems table and the copy included in release and Downloader/Update All artifacts.
- An ordinary upgrade using the unmodified shipped table must receive the corrected mapping automatically.
- Preserve user-edited tables under the existing custom-configuration contract.

## Acceptance checks

- With only `_Console/NGPC_YYYYMMDD.rbf` installed, Neo Geo Pocket and Neo Geo Pocket Color appear under Console, or Handheld when enabled.
- A `.ngp` game from the legacy `games/NGP` folder launches through `_Console/NGPC`.
- `.ngp`, `.ngc` and `.npc` examples from `games/NGPC` launch through `_Console/NGPC` at index `1`.
- Generated temporary MGLs and new Degauss favourites contain `_Console/NGPC` and no JT-specific set name.
- With only the appropriate JT core installed, the existing legacy launch still works.
- With both current and JT cores installed, the current core is selected by default.
- Existing metadata, artwork, favourites and saved settings remain attached to the same logical systems.
- Test the corrected launches on MiSTer with at least one monochrome and one colour game.

### LINKED ISSUE #93: Let each system choose among compatible installed cores
## Summary

Add a per-system **Launch Core** selector for choosing among compatible installed core families declared by Degauss.

This is separate from the existing **Core Version** selector:

- **Launch Core** chooses the compatible core family, such as current `NGPC` or legacy `JTNGPC`.
- **Core Version** continues choosing Standard, RetroAchievements or an Unstable build of the selected family.

The selector must not be an unrestricted browser that can assign any installed RBF to any system. A core may require a particular set name or launch protocol, so only combinations declared compatible in the systems table are safe to offer.

## Why this is needed

`systems.toml` currently gives each logical system one system-level `rbf` and optional `setname`. Users must edit that file manually when two compatible core families exist or a newer core replaces an older mapping.

Issue #92 is a concrete example: Neo Geo Pocket libraries can use the current `_Console/NGPC` core or retain Jotego's older cores as compatible fallbacks. Editing `systems.toml` is unsuitable as the normal user interface and also turns the file into a preserved custom configuration that will not automatically receive later shipped corrections.

Issue #87 is related but distinct. It selects a required core from the selected file format. This issue selects one compatible core profile for the whole logical system. Format-specific routing must continue to take precedence where declared.

The optional Cores browser in #85 is also separate: it launches a selected core directly and must not change how games in a system launch.

## User interface

Add **Launch Core** to the System group in Actions whenever the current system declares more than one compatible core profile. In Favourites, it applies to the original system of the selected favourite, following the existing Core Version behaviour.

Selecting the row opens a simple list, for example:

```text
Launch Core

Automatic (NGPC)
NGPC
JTNGPC (Legacy)
```

- **Automatic** is the default and persists nothing.
- Its label shows the compatible installed profile currently resolved by Degauss.
- Explicit profiles pin that choice.
- Show only declared compatible profiles that are installed, plus an explicitly saved profile that has since become unavailable.
- An unavailable saved profile remains visible as unavailable and fails clearly at launch. Do not silently change an explicit choice to another core.
- Selecting Automatic removes the saved override and always remains possible even if core enumeration fails.

Systems with no declared alternative retain the existing Actions menu without an empty or useless Launch Core row.

## Compatible core profiles

Extend a system entry with an optional ordered list of compatible core profiles. The existing system-level `rbf` and `setname` remain the first and default profile, so every existing table keeps its current meaning. Each additional profile needs:

- a stable profile ID;
- a user-facing label;
- an RBF reference;
- an optional set name.

The existing system extensions and launch rules remain shared. Only alternate cores compatible with those launch rules belong in this first implementation. Do not expose a profile that needs different file indexes, methods or delays. Format-specific core selection remains the responsibility of launch rules from #87.

Automatic resolution tries the default system-level profile first, then declared compatible profiles in table order, and selects the first installed one. This preserves current-core preference while allowing an older compatible installation to continue working. If none is installed, retain the existing missing-core error.

## Persistence and launch resolution

Store explicit choices in `settings.toml`, keyed by the existing stable system ID and containing only the profile ID. Do not rewrite `systems.toml`.

Resolve launch configuration in this order:

1. Required format-specific core routing from the matching launch rule, if present.
2. Explicit per-system Launch Core profile, otherwise Automatic profile resolution.
3. Existing per-system Core Version choice or global Standard/RA preference for that selected family.

The effective RBF and set name must be used consistently for normal launches, ZIP-member launches and Degauss-created favourites. Existing favourites must resolve through their original logical system without being recreated.

Changing Launch Core must take effect immediately. It must not rebuild the game list, modify cache identity, rescan game folders or rewrite metadata.

## Performance

- Check only the small set of declared profile paths against the shallow installed-core discovery Degauss already performs.
- Opening the selector must not enumerate game folders, ZIP files, artwork or metadata.
- No new startup traversal is permitted.
- The future cached catalogue from #85 may be shared, but this feature must not depend on #85.

## Backward compatibility

- A systems table without compatible profiles behaves exactly as it does now.
- A settings file without a Launch Core choice behaves exactly as it does now.
- Existing `core_choices` values for Standard, RA and Unstable remain valid and retain their meaning.
- Existing system IDs, game caches, metadata, artwork, favourites, hidden state and saved views remain unchanged.
- Existing custom `systems.toml` files continue to parse without modification.
- Unknown or removed saved profile IDs produce a clear unavailable choice and can always be reset to Automatic.
- Do not require a migration, reset, full index or targeted system rebuild.

## Out of scope

- Assigning an arbitrary installed core that Degauss has not declared compatible.
- Editing game folders, extensions, set names, indexes, methods or delays from the controller UI.
- Per-game core overrides.
- Changing what the direct Cores browser launches.

These require a substantially broader system editor and should not be hidden inside this selector.

## Acceptance checks

- A system without alternative profiles has no Launch Core row and behaves byte-for-byte as before.
- A system with current and legacy profiles shows Automatic plus both installed choices.
- With only the current profile installed, Automatic selects it.
- With only the declared legacy profile installed, Automatic selects the legacy fallback.
- When both are installed, Automatic selects the system-level current profile.
- An explicit choice survives restart and overrides Automatic ordering.
- Removing an explicitly selected core shows it as unavailable and produces a clear launch error without selecting another profile.
- Selecting Automatic removes the persisted override.
- Standard, RA and Unstable Core Version selection resolves against the chosen family without changing the Launch Core choice.
- A favourite uses the Launch Core setting of its original system, and changing the setting does not require recreating the favourite.
- Format-specific routing from #87 continues to win where present.
- Changing Launch Core causes no game-library scan, index rewrite or cache migration.
- Test the complete flow on MiSTer with the current NGPC core and one legacy JT core, including a normal game and a Degauss-created favourite.

### LINKED ISSUE #94: Add an optional launchable MiSTerZine updates browser
# Add an optional launchable MiSTerZine updates browser

## Problem

MiSTerZine provides a current catalogue of MiSTer releases and updates, but a passive feed inside Degauss would still make users leave the frontend to find out whether an entry is relevant to their own installation.

Degauss should optionally match MiSTerZine entries to content already installed on the SD card and launch that local content directly. It must remain a browser and launcher, not an updater: it must never download or execute a remote core, MRA or script.

## Data source

Use MiSTerZine's documented on-device contract:

- `releases/meta.json` to determine whether the catalogue changed.
- `releases/data.json` for structured release records and launch identifiers.

Do not scrape the website or render remote HTML. RSS is useful for general feed readers, but `meta.json` and `data.json` are the appropriate primary source here because they contain the structured MRA paths, core identities and release information required for matching and launching.

Sources:

- [MiSTerZine data and feed contract](https://github.com/matijaerceg/misterzine/blob/main/README.md#L238-L252)
- [MiSTerZine on-device frontend](https://github.com/matijaerceg/misterzine-on-device)

### Opt-in entry

- Add `Show MiSTerZine` under the general Library options.
- Default it to `Off`, including when the setting is absent on an existing installation.
- When enabled, add `MiSTerZine` to the main menu alongside the other optional top-level entries.
- When disabled, do not perform MiSTerZine network requests, parsing, matching or startup work.

### Updates browser

- Show recent catalogue entries in a normal Degauss list, newest first.
- Identify the content category, catalogue date and local installation state without replacing or modifying normal game metadata.
- Use Degauss's existing layouts, theme, local artwork and normal missing-art state.
- Do not package or download MiSTerZine images.
- Provide these contextual actions:
  - `Refresh`
  - `Installed only`
- `Installed only` is a view filter, not a persisted change to the library or catalogue.

### Match installed content

Match against Degauss's existing indexed library data and installed-core catalogue. Opening MiSTerZine must not walk the full SD card or rebuild any system.

- For Arcade records, match the catalogue's exact relative `.mra` path to an indexed local MRA.
- For Console, Computer and Other records, match the catalogue core identity through the existing core-name normalization and resolution rules, including dated RBF filenames.
- Reuse the cached Cores catalogue from #85 if it exists when this is implemented. Otherwise retain only the minimal shallow installed-core inventory already discovered by Degauss. This issue must not introduce a second independent full-card scan.
- Reuse the existing deterministic core resolver rather than inventing a separate priority order for Standard, RA and Unstable cores.
- Do not guess from a title when an exact path or compatible core identity cannot be established.

The basic local states are:

- `Installed`: a safe local launch target was found.
- `Not installed`: no safe local target was found.
- `Version unknown`: the target is installed but the available catalogue and filename evidence cannot establish its version.

`Current` or `Update available` may be shown only when the catalogue's build information and local evidence support that conclusion. Do not infer either state from the catalogue publication date alone. Any hashing needed for undated cores must be bounded and cached, never repeated while rendering or navigating the list.

### Launch local content

- Pressing `A` on an installed Arcade entry launches the matched local `.mra`.
- Pressing `A` on an installed Console, Computer or Other entry launches the matched local `.rbf`.
- Use Degauss's existing launch outcome and exit-on-launch path.
- Degauss must not remain resident while the launched content runs.
- A missing, ambiguous or no-longer-existing target must not launch anything. Show an actionable message and allow the user to continue browsing.
- A non-installed entry remains informative but is not launchable.

An Arcade core update may affect several games. `data.json` exposes the individual Arcade records, so Degauss should present and launch the installed games rather than adding an abstract aggregate core-update row.

### Refresh, cache and offline behaviour

- Keep MiSTerZine data in a separate Degauss cache. Do not change system caches, `gamelist.xml`, Artwork Pack data or user metadata.
- On opening the section, display a valid cached catalogue immediately when available, then check `meta.json` away from the UI thread.
- Download and parse `data.json` only when the published catalogue identity changed or the user explicitly selects `Refresh`.
- Keep only the fields needed by the view and matching logic after parsing.
- If a refresh fails and a valid cache exists, keep it and show an explicit offline/error state with the cache date. Do not silently present stale data as current.
- If no cache exists, show the real retrieval or parsing error and allow `B` to return normally.
- Cancellation and navigation must remain responsive throughout fetching, parsing and matching.
- Do not add a daemon, periodic background service or work that continues after Degauss exits.

Use a fixed official HTTPS source, bounded response sizes, short connection and overall timeouts, and defensive parsing. Unknown future JSON fields must not break compatible records. Never render remote HTML or execute remote links or paths.

### Attribution

Show concise MiSTerZine attribution in the section or its information view and document the data source. The catalogue is published under CC BY 4.0; artwork retains its owners' rights and is therefore deliberately not redistributed by this feature.

## Backward compatibility

- Existing installations with no new setting behave exactly as before.
- Existing startup, indexing, game browsing, artwork, metadata and launching paths remain unchanged while the feature is disabled.
- Enabling or disabling the section requires no migration, reset or reindex.
- A corrupt or outdated MiSTerZine cache must not damage or invalidate any Degauss library cache.
- Returning from locally launched content follows the same behaviour as every existing Degauss launch.

## Performance requirements

- Zero MiSTerZine network, filesystem matching or parsing work during startup when disabled.
- No full-card scan or system rebuild when opening or refreshing the section.
- Fetching, parsing and matching run away from the UI thread with visible progress and working cancellation.
- Warm opening from a valid cache must be effectively immediate on MiSTer.
- The parsed in-memory representation must contain only the recent records and fields needed by this feature.

## Tests and acceptance

Automated tests should cover:

- Missing setting and explicit `Off` perform no MiSTerZine work.
- Catalogue parsing tolerates unknown fields and rejects malformed required launch identifiers safely.
- Exact Arcade MRA matching.
- Dated and undated core-name normalization.
- Installed, not-installed and version-unknown states.
- No launch for missing or ambiguous targets.
- Launch plans use only resolved local paths.
- Failed and cancelled refreshes preserve an existing valid cache and report their state.
- A changed `meta.json` identity refreshes `data.json`; an unchanged identity does not.
- Filtering to installed entries does not modify persisted library data.

On a real MiSTer, verify:

1. Existing startup and browsing with the option disabled.
2. First online open with no cache, including progress and cancellation.
3. Warm cached open.
4. Offline open with and without a cache.
5. Refresh with unchanged and changed catalogue metadata.
6. Launch and return for one installed Arcade MRA and one installed non-Arcade core.
7. Safe handling of an entry removed from the card after matching.
8. Repeated opening, filtering and refreshing without a reindex, UI stall or increasing memory use.

Record startup and cold/warm opening measurements on the same MiSTer and configuration. The disabled feature must not produce a measurable startup regression.

### LINKED ISSUE #97: Support direct launch from AmigaVision April 2026 friendly listings

### LINKED ISSUE #100: Show one-game ZIP archives directly as games
## Problem

Degauss opens a ZIP as a virtual folder whenever the current system does not launch the whole archive directly. This is required for multi-game ZIP libraries, but it adds an unnecessary navigation level to one-ROM-per-ZIP libraries:

```text
System/
  Game.zip/
    Game.rom
```

The person must open `Game`, then select the only game inside it. This is not specific to SNES or Genesis. It affects every system using Degauss's generic virtual-ZIP handling.

This is a follow-up to #32 and #48. It must not change the dedicated Neo Geo ROM-set behaviour implemented by #56.

## Required behaviour

- When a generic ZIP contains exactly one member supported by the current system, show that member directly as a normal game in the ZIP's containing folder.
- Apply the rule to every system using generic virtual-ZIP handling. Do not hard-code a list of systems or extensions.
- Count supported launchable members, not every archive entry. A ZIP containing one supported game plus a README, artwork or other unsupported files still contains one game.
- Apply the same rule when the sole supported game is inside one or more directories within the ZIP.
- Launch the exact existing native target, `archive.zip/path/to/game.ext`. Do not extract, rewrite or replace the archive.
- Preserve the sole member's existing effective title, metadata, artwork, favourite state and launch rules, including the existing single-game archive-level `gamelist.xml` compatibility.
- Keep the current virtual-folder behaviour when a ZIP contains more than one supported game.
- Keep existing whole-archive pass-through unchanged for systems whose configuration accepts `zip` directly.
- Keep recognised Neo Geo ROM-set ZIPs unchanged. They remain whole ROM sets identified through `romsets.xml` and passed to the Neo Geo loader as complete ZIPs.
- Keep malformed, unsupported and unreadable archive reporting unchanged.
- Do not flatten ordinary filesystem folders. This issue applies only to a generic ZIP with exactly one supported game.
- Reuse the archive directory read already required during indexing. Do not add a second archive scan, synchronous archive access while scrolling, or work to ordinary non-ZIP libraries.

## Backward compatibility

- Existing multi-game ZIP libraries retain their hierarchy and exact member identities.
- Existing Favourites targeting the sole member remain valid because their `archive.zip/member` target does not change.
- Existing metadata and artwork mappings continue to bind to the same member.
- Existing settings, system definitions, caches and state remain readable without migration or reset. A normal system rebuild applies the new presentation.
- If saved navigation points inside a ZIP that is now flattened, return to its containing folder and select the flattened game using the existing missing-place recovery behaviour.

## Acceptance checks

- [ ] A ZIP containing one supported game at its root appears as one directly launchable game, without an intermediate folder.
- [ ] A ZIP containing one supported game in a nested directory behaves identically.
- [ ] Unsupported companion files beside the sole game do not prevent flattening.
- [ ] A ZIP containing two supported games remains a virtual folder and both games remain independently launchable.
- [ ] A system configured to launch `zip` directly retains its current whole-archive behaviour.
- [ ] Recognised zipped Neo Geo ROM sets retain their current titles, artwork and whole-set launch targets.
- [ ] Existing single-game archive-level and exact-member Gamelist metadata continue to resolve correctly.
- [ ] An existing Favourite for the sole member survives rebuild and restart and launches the same target.
- [ ] Search and return position treat the flattened row as the same game member, not as a new archive identity.
- [ ] Malformed and unsupported ZIP handling remains unchanged.
- [ ] Real MiSTer tests launch representative one-game ZIPs through SNES and Genesis, while a small multi-game ZIP still opens as a folder.
- [ ] On-device indexing comparison confirms no material regression for ordinary folders or a library of one-game ZIPs.

## References

- #32
- #48
- #56
- [Current generic ZIP-folder rule](https://github.com/giancarloerra/Degauss/blob/v0.7.1/src/browse.rs#L668-L675)
- [Current single-game ZIP presentation](https://github.com/giancarloerra/Degauss/blob/v0.7.1/docs/zip-libraries.md#L1-L3)
- [Current Neo Geo ROM-set distinction](https://github.com/giancarloerra/Degauss/blob/v0.7.1/src/neogeo.rs#L1-L10)

giancarloerra:
## What this changes

- Adds optional metadata filters, Last Played, Cores and MiSTerZine browsers.
- Adds display-only game-name formatting and optional folder brackets.
- Adds current Neo Geo Pocket core mappings and per-system launch-core selection.
- Adds current AmigaVision mappings and direct browsing of one-game ZIP archives.
- Updates the README with the approved 0.8.0 screenshots.
- Combines the completed 0.8.0 work and sets the version to 0.8.0.

## Why

This combines the related completed changes into one release integration pull request.

Closes #83
Closes #84
Closes #85
Closes #86
Closes #92
Closes #93
Closes #94
Closes #97
Closes #100

Supersedes #88, #89, #90, #91, #95, #96, #98, #99 and #102.

## How it was checked

- [x] `cargo test`
- [x] `cargo fmt --check`
- [x] Run on a real MiSTer

The RA release compatibility check and all migration rehearsals also pass.

## Licensing

- [x] I agree

## Summary by CodeRabbit

- **New Features**
  - Added optional browsing for installed cores and MiSTerZine releases.
  - Added metadata filters, configurable game-name display, folder bracket settings and “Last Played” history.
  - Added support for compatible core families, including Neo Geo Pocket and Neo Geo Pocket Color.
  - Added direct launching for cores and MiSTerZine items.
  - Single-game ZIP archives now appear directly as games, including nested archives.
  - Added AmigaVision catalogue support and verified launch-name mappings.

- **Documentation**
  - Updated the README screenshot gallery and archive-library guidance.


giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

PR #109, opened 2026-09-18
Respect format-specific cores during launch

### LINKED ISSUE #107: Respect format-specific core overrides during interactive launch
## Regression

Issue #87 and PR #105 added the required format-specific mapping:

- `.bin` uses the legacy `GnW` core.
- `.gnw` uses the current `GameAndWatch` core.

The shipped table contains that mapping, but a system with only the current `GameAndWatch` core installed still rejects a `.gnw` launch with:

> launch core unsupported: no declared compatible core is installed for Game & Watch

The interactive launch path performs a system-level core check before it builds the launch plan. The Game & Watch system-level core remains `GnW`, so this check fails before the `.gnw` rule can select `GameAndWatch`: https://github.com/giancarloerra/Degauss/blob/7cb91cefa87322a59a54beadd299c9aa2805dc9f/src/app.rs#L8027-L8072

The planner already applies and validates the matching rule-level core correctly: https://github.com/giancarloerra/Degauss/blob/7cb91cefa87322a59a54beadd299c9aa2805dc9f/src/launch.rs#L736-L753

The shipped Game & Watch rules are here: https://github.com/giancarloerra/Degauss/blob/7cb91cefa87322a59a54beadd299c9aa2805dc9f/assets/systems.toml#L651-L670

## Required behaviour

- A `.gnw` game launches when `GameAndWatch` is installed, without requiring the legacy `GnW` core.
- A `.bin` game continues to launch through `GnW`.
- A fixed rule-level core is not replaced by the system's Core Version or Launch Core selection.
- A missing required core remains a visible error before Degauss exits.
- Existing self-describing `.mgl` files remain unchanged.

## Smallest correction

Remove the earlier system-only preflight and let the existing launch planner validate the effective core after the matching launch rule has been selected. The planner already returns its error to the same on-screen message path before any launch is executed, so a second core check is unnecessary and can disagree with the actual plan.

Do not add another Game & Watch system, change the shipped format mapping, or require a library rebuild or settings migration.

## Compatibility

- Keep the existing `GameNWatch` system identity, folders, metadata, artwork and cache association.
- Keep legacy `.bin` behaviour.
- Keep Standard, RetroAchievements and Unstable selection unchanged for systems without a fixed rule-level core.
- Keep Degauss-created favourites and supported ZIP members using the same effective format-specific core.

## Acceptance checks

- With only `GameAndWatch` installed, launch a current `.gnw` package from the normal Game & Watch list.
- With only `GnW` installed, launch a legacy `.bin` game.
- With both cores installed, confirm each format still selects its declared core.
- With `GameAndWatch` absent, confirm `.gnw` is rejected before handover with a specific missing-core error.
- Repeat `.gnw` through a Degauss-created favourite and from a supported ZIP library.
- Add an application-path regression test that reaches the interactive launch handler, because planner-only tests do not cover the failing preflight.
- Confirm a normal system without a rule-level `rbf` still rejects a genuinely missing selected core before exit.

giancarloerra:
## What this changes

Interactive launches now validate the core selected by the matching file-format rule instead of rejecting the system's default core first.

## Why

Closes #107.

## How it was checked

- [x] `cargo test`
- [x] `cargo fmt -- --check`
- [ ] Run on a real MiSTer

## Licensing

By opening this pull request I agree that my contribution is licensed under the [PolyForm Noncommercial License 1.0.0](../LICENSE), the same licence as the rest of Degauss, and that I have the right to license it that way.

Changes to `MiSTer_Degauss` do not belong here: it is a separate GPLv3 program in its own repository, [Degauss-Main](https://github.com/giancarloerra/Degauss-Main).

- [x] I agree

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved file-launch validation to select the appropriate core based on the file format.
  - Corrected Game & Watch launches to use the proper emulator core.
  - Added clearer launch errors when the required format-specific core is unavailable.
  - Preserved the existing dedicated preflight handling for AmigaVision launches.


giancarloerra:
@coderabbitai review

giancarloerra:
Closed by #113.

PR #113, opened 2026-09-18
Integrate TATE, launch-core, CD32 and MiSTerZine updates

### LINKED ISSUE #41: Add CRT TATE support with frontend rotation and portrait layouts
## Problem

Degauss needs to remain upright and usable on a physically rotated CRT or monitor. General CRT support does not establish TATE support: the frontend must rotate, lay out its content for portrait dimensions, and preserve readable controls and correct artwork proportions.

## Current implementation and feasibility

Analysis baseline: public commit [`150528c`](https://github.com/giancarloerra/Degauss/tree/150528c6295ec0264b88b0e09065334768d79e0c).

- [`src/surface.rs`](https://github.com/giancarloerra/Degauss/blob/150528c6295ec0264b88b0e09065334768d79e0c/src/surface.rs) owns physical framebuffer geometry, stride, format and mapping. Degauss does not change its video mode.
- [`src/render.rs`](https://github.com/giancarloerra/Degauss/blob/150528c6295ec0264b88b0e09065334768d79e0c/src/render.rs) uses Slint's line renderer, with direct and staged presentation and partial repainting. There is no frontend rotation control.
- [`src/main.rs`](https://github.com/giancarloerra/Degauss/blob/150528c6295ec0264b88b0e09065334768d79e0c/src/main.rs) sizes the window from the framebuffer. [`src/app.rs`](https://github.com/giancarloerra/Degauss/blob/150528c6295ec0264b88b0e09065334768d79e0c/src/app.rs) calculates layout geometry, currently including a side-by-side Details view, fixed two-column Multi List and minimum grid-column counts.
- The locked Slint renderer is 1.17.1. Its existing `SoftwareRenderer::set_rendering_rotation` supports rotation; `render_by_line` reports dirty regions in physical coordinates after rotation. Input remains in logical coordinates. Prefer this API over a second custom rotation/copy implementation. See the [official renderer API](https://docs.slint.dev/latest/docs/rust/slint/platform/software_renderer/struct.SoftwareRenderer).
- MiSTer's [`osd_rotate`](https://github.com/MiSTer-devel/Main_MiSTer/blob/master/MiSTer.ini#L42) is documented as rotating its OSD. It is not a documented Degauss framebuffer-rotation contract and should not be assumed to provide this feature.

The rotation primitive is relatively small. The larger work is making every screen usable at portrait CRT resolutions and verifying repainting, pixel aspect and launch/return on hardware. This is broader than an Options-menu reorganisation.

## Proposed behaviour

1. Add one global **Screen Rotation** setting: **Off**, **90° Clockwise**, **90° Counterclockwise**. Off is the default for existing installations. Labels describe the rotation applied to the interface, not the direction the user turned the monitor.
2. Apply it to the entire Degauss interface: startup, browsing, every menu/dialog, artwork, logos, keyboard, progress screens, theme editor and screensaver. It is independent of per-folder view choices and game/core rotation.
3. Use explicit selection. Do not infer orientation from the selected game, MRA, system, MiSTer OSD setting or framebuffer `rotate` field without a verified integration contract. Do not change MiSTer.ini, core settings, video timings or controller configuration.
4. Preview changes immediately. Ask **Keep This Rotation?** with A to keep, B to revert and a 15-second timeout to revert. Store only the confirmed choice; cancel, timeout or exit during preview restores the previous orientation. Display a settings-write failure with its cause and restore the previous confirmed choice if the new one cannot be saved. Returning to Off must work from the same menu.
5. Up and Down continue to move visually up and down; existing horizontal-navigation choices keep their meanings. Do not rotate controller button assignments. Any coordinate-based input path must be checked against Slint's logical-coordinate contract.
6. A single framebuffer has one Degauss orientation. Simultaneously connected displays show that same orientation; independently rotating HDMI and analog output is outside this feature.

## Geometry and presentation design

- Keep physical framebuffer width, height, stride, pixel format and mapping unchanged. For a quarter-turn, give the app and Slint a logical width equal to physical height and logical height equal to physical width. Example: physical 352×240 becomes a 240×352 logical interface.
- Use Slint's rendering rotation for both RGB565 and XRGB8888, in both existing presentation modes. Keep staging allocation and destination addressing in physical coordinates. Do not rotate dirty rectangles a second time.
- Centralise orientation and logical-dimension calculation so interactive rendering, startup, screenshots, benchmarks and framebuffer self-tests cannot disagree.
- On any orientation change or revert, resize/recompute the window and UI, clamp list selections/scroll offsets, invalidate geometry-dependent artwork sizing and force a complete repaint. Retain the current folder, selection, active theme and effective view.
- Make portrait sizing responsive to available width as well as height. Merely swapping dimensions can produce large text in a narrow column.
- Proposed portrait Details arrangement: artwork and metadata above a full-width game list. Maintain readable list space when artwork is missing or disabled. Adapt Tiled, Carousel, List, Multi List and Gallery to the available width; a layout may reduce its column count without changing the saved view name. Do not silently select another view.
- Audit plain menus, labels/values, help text, confirmation messages, keyboards, theme-editor controls and progress/cancellation controls for clipping and overlapping at narrow widths. Long titles may use the existing marquee; essential actions and current setting values must remain readable.

## Pixel aspect and screen adjustment

`ArtworkScale::horizontal` currently derives correction from framebuffer aspect and the chosen 4:3 or 16:9 display. Quarter-turn rotation changes the logical display aspect: a physically rotated 4:3 screen is 3:4, and 16:9 becomes 9:16. Derive correction in logical coordinates with the reciprocal physical aspect. Do not simply feed swapped dimensions into the existing landscape formula.

Preserve the existing Artwork Scale choices and Off-mode results. Apply correction once to game artwork; preserve the existing policy for logos and screensaver images while rotating them with the interface. Missing-image and missing-metadata behaviour remains unchanged.

Edge margins and position controls should operate along the visible interface's horizontal and vertical axes. Keep the existing stored X/Y values and their Off-mode meaning; enabling TATE does not rewrite or transpose saved values. Explain the visible-axis meaning in the controls' help.

## Implementation sequence

1. Prove the Slint API with asymmetric corner markers through the actual line providers, both formats and presentation modes, including partial redraw and padded stride. Establish clockwise/counterclockwise mapping before UI work.
2. Add the optional persisted orientation and central logical/physical geometry boundary. An absent value means Off. An invalid explicit value must produce an actionable configuration error, not a silently changed orientation. Preserve existing settings and custom views.
3. Add interactive preview, confirmation, timeout/revert and startup application before the first visible Degauss frame.
4. Adapt portrait layouts and auxiliary screens, artwork correction and visible-axis screen adjustments.
5. Add a non-persisting `--rotation off|cw|ccw` override for `--render`, `--bench` and `--selftest`; explicit CLI choice takes precedence over saved orientation. `--geometry` remains physical output dimensions and existing `--layout` precedence is preserved. Report physical dimensions and rotation clearly.
6. Integrate the control into Display if the Options reorganisation has landed; otherwise add it to the current display group. Neither issue is a prerequisite for the other.

## Acceptance and test plan

| Area | Required evidence |
| --- | --- |
| Rendering correctness | Distinct corner/edge markers prove both directions; both formats and presentation modes produce equivalent pixels. Padded stride and buffer guards remain intact. Partial redraw updates the correct physical rectangle; toggling/reverting never leaves old pixels. |
| Portrait UI | Render and inspect all six views and every auxiliary screen at physical 352×240 and 640×480 with each quarter-turn, plus a larger HDMI geometry. Test long labels, empty lists, missing art, custom views, all bundled fonts and overscan extremes. Every action remains reachable. |
| Landscape compatibility | Off-mode captures and workflows remain unchanged. Old settings fixtures load without migration; toggling orientation does not alter custom views, sources, favourites or launch selection. |
| Persistence and recovery | Confirm, cancel, timeout, repeated changes, save failure and restart. Confirmed orientation survives reopening; unconfirmed preview does not. CLI overrides do not alter saved settings. |
| Artwork and placement | Synthetic circles/squares and known-aspect images verify 4:3/16:9 correction, both rotations, margins and shifts. Physical-display inspection confirms proportions that a raw framebuffer screenshot cannot establish. |
| Target execution | Run the real MiSTer executable through its installed launcher and actual framebuffer, test menus and scrolling, screensaver entry/wake, one game launch and frontend return. Verify the selected orientation persists and game video settings were not changed. |
| Performance | Compare Off and both rotations on the same MiSTer using existing benchmark/self-test instrumentation. Record render/copy timings, input responsiveness and full/partial repaint behaviour; investigate regressions before declaring support. |

Run relevant format/compiler/lint checks and the complete applicable test suite after implementation. Automated framebuffer captures can validate pixel orientation without physically moving a display. An actual safely positioned TATE CRT is still required to validate readability, physical aspect, overscan and the analog result; do not claim that host renders establish those properties. Verify each available output path separately and state any untested hardware path explicitly.

## Scope and open validation

This is a proposed implementation, not a working TATE feature or a hardware-validation result. The Slint API removes the need to invent a renderer but still needs integration testing with Degauss's presenter. Support for a custom MiSTer video configuration must be established from its actual framebuffer/output behaviour. Automatic orientation, independent output rotation, game-video changes and unrelated display/colour bugs are outside scope.

### LINKED ISSUE #107: Respect format-specific core overrides during interactive launch

### LINKED ISSUE #108: Launch standalone AmigaVision CD32 games through their supplied MGLs
## Problem

The standalone AmigaVision CD32 package does not launch CD32 games from raw CHDs alone. It supplies one MGL and one generated Minimig configuration per title, plus the ROM, boot HDFs and save HDF required by those configurations.

The package's MGLs launch `_computer/minimig` with a title-specific `setname`, which makes MiSTer load the matching generated configuration: https://github.com/amigavision/AmigaVision/blob/c3e2693a5062d3484f548ed21a2ae5fab991847f/cd32/make_cd32_mgl_cfg.py#L255-L262

Those configurations mount the CD32 ROM, the appropriate boot HDF, the save HDF and the exact CHD: https://github.com/amigavision/AmigaVision/blob/c3e2693a5062d3484f548ed21a2ae5fab991847f/cd32/make_cd32_mgl_cfg.py#L442-L476

Degauss currently indexes `games/AmigaCD32` and generates a generic temporary MGL that mounts only the selected CHD under Minimig: https://github.com/giancarloerra/Degauss/blob/7cb91cefa87322a59a54beadd299c9aa2805dc9f/assets/systems.toml#L50-L61 and https://github.com/giancarloerra/Degauss/blob/7cb91cefa87322a59a54beadd299c9aa2805dc9f/src/launch.rs#L827-L855

This bypasses the supplied title configuration, so the native MiSTer entry can work while the same game launched from Degauss cannot boot correctly.

## Required behaviour

For a selected `.chd` in `games/AmigaCD32`:

1. Derive the storage root that contains that game.
2. Look for an exact same-stem MGL at `_Console/_Amiga CD32 Games/<title>.mgl` on that same storage root.
3. If it exists, launch that supplied MGL unchanged instead of generating Degauss's generic temporary MGL.
4. If it does not exist, preserve the current generic CD32 launch behaviour.

This must work for the package's SD, USB and NAS layouts. It must not scan all 1,935 supplied MGLs on every launch or add them as duplicate game rows. One exact path check for the selected game is sufficient.

## Compatibility

- Keep the existing `AmigaCD32` system identity, `games/AmigaCD32` library, list order, names, metadata, artwork and cache format.
- Keep `.cue`, `.iso`, direct `.mgl` and non-AmigaVision CD32 behaviour unchanged.
- If no matching package MGL is installed, use the existing launch path exactly as before.
- Degauss-created favourites that point to a CD32 CHD must apply the same matching-MGL rule when launched, without recreating or migrating them.
- Do not require a rebuild, reset or configuration edit.

## Acceptance checks

- Install the current standalone AmigaVision CD32 package and one correctly named CHD.
- Confirm the title launches through MiSTer's native `Console -> Amiga CD32 Games` entry.
- Launch the same title from Degauss and verify that Degauss hands over the package's same-stem MGL rather than a generated generic MGL.
- Test one title using the normal generated setup and one title whose generated configuration selects a different boot HDF or per-game setting.
- Repeat one launch through a Degauss-created favourite.
- Test the SD layout and at least one external-storage layout.
- Remove or rename the matching MGL and confirm the existing generic path remains unchanged.
- Confirm indexing time and the visible game list are unchanged.

## Upstream package contract

The standalone package documents `games/AmigaCD32` for CHDs and `Console -> Amiga CD32 Games` for launching the supplied MGLs: https://amiga.vision/cd32

### LINKED ISSUE #111: Show MiSTerZine updates for all installed cores
Rename the optional feature to **MiSTerZine Updates** and show every matching release that is installed and launchable on the MiSTer, regardless of its source or developer.

## Behaviour

- Preserve existing launch and version-comparison behaviour.
- Preserve the existing opt-in setting and stored value. Do not rename the configuration key or require a migration or reset.
- Keep refresh, caching, cancellation, filtering and failure reporting responsive.
- Do not perform network checks during startup. Refresh only when MiSTerZine Updates is opened or explicitly refreshed.

## Acceptance criteria

- A user who already enabled MiSTerZine remains enabled after updating.
- Installed and launchable content from every source appears and launches as before.
- An installed core outside the standard distribution remains visible and can report an update.
- Uninstalled or non-launchable content does not appear.
- Opening the feature, refreshing it, cancelling it and using a valid cached result remain responsive.

giancarloerra:
## What this changes

- Adds opt-in CRT TATE rotation.
- Respects format-specific core choices during launch.
- Uses supplied AmigaVision CD32 launchers.
- Refines MiSTerZine Updates for installed, launchable content.

## Why

Closes #41.
Closes #107.
Closes #108.
Closes #111.

## How it was checked

- [x] `cargo test`
- [x] `cargo fmt --check`
- [ ] Run on a real MiSTer

## Licensing

By opening this pull request I agree that my contribution is licensed under the [PolyForm Noncommercial License 1.0.0](../LICENSE), the same licence as the rest of Degauss, and that I have the right to license it that way.

Changes to `MiSTer_Degauss` do not belong here: it is a separate GPLv3 program in its own repository, [Degauss-Main](https://github.com/giancarloerra/Degauss-Main).

- [x] I agree

## Summary by CodeRabbit

- **New Features**
  - Added clockwise and counterclockwise screen rotation, including preview, cancellation, timeout, persistence, and command-line override support.
  - Added portrait-friendly layouts for browsing, details, artwork, pickers, and settings.
  - Added automatic Amiga CD32 launcher discovery for supported game files.

- **Improvements**
  - MiSTerZine Updates now focuses on launchable locally installed content and removes the Installed Only action.
  - Updated labels, descriptions, screenshots, and help text to reflect the new behaviour.


giancarloerra:
@coderabbitai full review

giancarloerra:
@coderabbitai full review

